PR #6, opened 2019-09-14
Define a minimum window size when supported

Calinou:
This prevents UI elements from collapsing when resizing the window.


OverloadedOrama:
Great idea, thanks for the contribution!

PR #14, opened 2019-10-25
Remove trailing whitespace and ensure newlines at end of files

Calinou:
This makes for cleaner VCS diffs. It's usually considered good practice to avoid trailing whitespace in files, and to make sure all files end with a single blank line.

PS: You can enable **Text Editor > Files > Trim Trailing Whitespace On Save** in the Editor Settings to do this automatically in the script editor :slightly_smiling_face:

**Should be merged *after* #13.**


OverloadedOrama:
Thanks, that was something I wasn't aware of. I enabled **Trim Trailing Whitespace On Save** in the Editor Settings :)

PR #24, opened 2019-11-22
Fixed flipping sprites not updating

danielnaoexiste:
Fixed the flip horizontal and vertical not updating until another draw action happens.


OverloadedOrama:
Hey there! Flipping a sprite calls canvas' handle_redo(), which in turn calls redo() from Global, which updates the canvas' texture. Are there times when the update is not happening? Because I can't replicate the issue.

danielnaoexiste:
It was actually happening but i can't seem to replicate it now for some reason. You can just close this pull request i guess

Issue #10, opened 2019-09-27
Weird crash when using fill bucket


OverloadedOrama:
I've noticed a very weird bug that leads to a crush when using the fill bucket tool.
Steps to reproduce:

1. Select the bucket tool
2. Change the selected color
3. Add it to the preset colors (the palette)
4. Change the color again
5. Click on the canvas to fill
6. The pixels get filled as normal, but Pixelorama crashes.

I haven't discovered why this happens, and it doesn't seem like a problem with my code, because the colors get filled, which means that it's not the flood fill algorithm's fault. This also means that the frame gets drawn, but the next one leads to a crush. I'm starting to think whether it's a problem with Godot and the Color Picker itself, but if any of you has any other ideas, let me know!

![fill bucket CRUSH](https://user-images.githubusercontent.com/35376950/65790454-575be780-e168-11e9-8936-af8cced91a24.gif)

Flairieve:
I have worked on a graphics editor plugin before on Godot. Most of the time a crash like this would have to do with something in gdscript looping constantly. Did you try debugging it by making a print statement in the bucket fill function?

Calinou:
Remember that's there's a stack limit in functions, so you need to make sure recursive functions don't run too deep. The default call stack limit is 1024 (it can be adjusted in the Project Settings).

OverloadedOrama:
> I have worked on a graphics editor plugin before on Godot. Most of the time a crash like this would have to do with something in gdscript looping constantly. Did you try debugging it by making a print statement in the bucket fill function?

I did try and do that, it prints as it should. The bucket fill function is executing as it should, otherwise it would crash before it gets filled with color. Which means, the crash happens after the function is executed, therefore, the problem is not in the function itself. Which is exactly why this is such a weird issue.

Flairieve:
Maybe try to print each line that comes after calling the bucket fill function to see where it crashes.

OverloadedOrama:
> Maybe try to print each line that comes after calling the bucket fill function to see where it crashes.

I tried that too. I printed something at the end of _process and it worked. Then, I tried printing both at the end and at the start of _process. The line at the end was printed, but it stopped there, and the line at the start wasn't printed after that. This should mean that the crush happens at the start of the next frame, and not at the current frame. Right before any code manages to get executed.

novhack:
[...]

PR #31, opened 2019-11-30
Another attemt to workarround flood_fill bug

azagaya:
This is another attempt to workarround the bug reported in #10, although it doesn't address the real cause of it. I think there may be a bug in the engine itself, given the strage steps that cause the problem.


OverloadedOrama:
I also think it's a bug in the engine. But I like this workaround so I'll go ahead and merge it!

Issue #85, opened 2019-12-22
High CPU Usage


Martin1991zab:
Hello,

i notified that the Pixelorama eats 2 cpu cores while in idle completely.

I have a laptop with Intel Core i7 7600U (2 Cores - 4 Threads) running Fedora 31.
Pixelorama is permanently at 50-60% in htop, for comparison the godot editor at only 5%.

Calinou:
This is due to Pixelorama not using low-processor usage mode. We could enable it, but something is causing Pixelorama to redraw constantly while it's enabled (likely something in `_process`).

OverloadedOrama:
So the cause of this might be that `update()` (which is used to update the `_draw()` method in CanvasItem objects) is called in `_process()`? I could try and optimize this if this is the case.

Calinou:
@OverloadedOrama That could be try. Try to enable low processor usage mode in the Project Settings and monitor the CPU usage.

OverloadedOrama:
I enabled it in c840bed941cfb73beb16a6568abfd5abdf17be55. I think I notice less CPU usage now, but the different is not very big. Might make a bigger difference on low-end software though.

Martin1991zab:
Well yes ... roughly 5% less with my laptop, but still 45% - 50% in idle. When i start to paint something we have 60% again.
Tested with: d3ff1e984b20d403fe1f4fbc0035311f6b61d08c

OverloadedOrama:
Hmm I see... I'll try optimizing how often `_draw()` methods are being called then, at some point.

OverloadedOrama:
CPU usage should be less now when idle, since `update()` methods aren't being called all the time, in Canvas.gd in 3a20b3e073339c49ab2e7cfcdcfe7541cb2bf755 and the rulers in 037a797afc7fdc4e6a3a2d6f74a49ec1f584d732. The canvas could use a few more optimizations however, because it's being re-drawn every time it detects input.

Martin1991zab:
Tested with: 61169f86bab6e70a7a178d37725326efb46f5078 on Godot master (f90e8c41)
Result: down to 40-45% on Godot beta - idle and drawing

Martin1991zab:
080676ac8342b63a47088ec2231ef513f157a882 is not really a change

it seams like the canvas is not that big problem.
is set a print in the draw function: 

- print(String(OS.get_ticks_msec()) + ": redraw canvas")
- the first line is from opening the program
[...]

PR #95, opened 2019-12-27
Don't call `update()` in the canvas on non-mouse input events

Calinou:
This partially addresses #85.


OverloadedOrama:
I just realized that this stops straight lines (with Shift) from working.

Calinou:
Sorry, I forgot to check that :slightly_frowning_face: Feel free to revert it, or make it so key inputs cause an update if `KEY_SHIFT` is being held or released.

OverloadedOrama:
Fixed in 42e3aec56136d10be88a951a994b4a094b42fa58 :)

PR #103, opened 2020-01-01
Make the Splash Screen popup optional and German localization UPDATE

Schweini07:


Schweini07:
It's very strange but somehow you can't draw anymore. Please don't merge now.

Schweini07:
Now it should work

PR #119, opened 2020-01-11
Holding Ctrl when using the pencil tool makes…

GreenBeakCrow:
…it pick the color instead of drawing.
This is GIMP behaviour and I find it very useful. The implementation might not be optimal, but I thought I offer you the change nonetheless. You can take it just as an inspiration.

### DIFF
--- a/Scripts/Canvas.gd
+++ b/Scripts/Canvas.gd
@@ -191,6 +191,9 @@ func _input(event : InputEvent) -> void:
 			if previous_action != "None" && previous_action != "RectSelect" && current_action != "ColorPicker":
 				handle_redo("Draw")
 
+	if current_action == "Pencil" and Input.is_action_pressed("ctrl"):
+		current_action = "ColorPicker"
+
 	match current_action: # Handle current tool
 		"Pencil":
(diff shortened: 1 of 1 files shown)


OverloadedOrama:
Not sure if I can merge it, as it will break the constrained lines functionality, which needs both Shift (to make a straight line) and Ctrl to work.

You can use the shortcut "O" to use the color picker, and then "P" for the pencil anyway, or you can map the color picker to your right button and have it pick for the left color. This was added some commits ago and it will be available in v0.6.1. I'm not sure if pressing Ctrl to pick colors is really needed here.

axredneck:
It would be good as an option

Schweini07:
This PR is pretty old, still relevant?
Also, @novhack made a PR with which you can change the shortcuts, so now you can (probably) do the same.

OverloadedOrama:
I think I'll close this since it doesn't seem it's really needed, and Ctrl is used on the pencil tool to constrain angles, along with Shift. The ability to change shortcuts has been added but won't work for this case, since it only lets you change tool shortcuts, for now, and Ctrl is considered already assigned. This could be changed in the future but we'll see. For now I'll close this issue.

PR #123, opened 2020-01-13
added option to hide tool icons next to cursor

### LINKED ISSUE #122: Add option to disable tool icons next to cursor

haonkrub:
Fixes #122 

implemented using checkboxes  in the the left and right tool options panels

image

translations are done using google translate, so they should be checked by translators


OverloadedOrama:
Thank you!

PR #124, opened 2020-01-14
Delay the splash screen popup so it shows properly centered

YeldhamDev:
When the window adjusts its size when the program it's launched, the splash screen can be off-centered. This commit delays the popup to insure it will be centered.


OverloadedOrama:
Awesome, thank you!

Issue #42, opened 2019-12-15
Feature: Rotsprite


msmshazan:
Implement rotsprite for sprite rotation 
See https://forums.sonicretro.org/index.php?threads/sprite-rotation-utility.8848/ for implementation details

OverloadedOrama:
Hello. The link you attached is old and the download link doesn't seem to work, so I'm not sure if this can be used. I'll do some more research to see if I can find another algorithm for rotating sprites, to avoid re-inventing the wheel.

msmshazan:
Oh sorry here's the updated link with the utility http://info.sonicretro.org/RotSprite

msmshazan:
I also found the Wikipedia article describing it https://en.m.wikipedia.org/wiki/Pixel-art_scaling_algorithms#RotSprite

OverloadedOrama:
I took a look. This looks very interesting, but a little advanced for now. I'll definitely give it a try in the future though!

PR #143, opened 2020-02-03
Added 2 rotation algorithms and popup for rotating current layer

azagaya:
Added a custom rotation algorithm (i named it rotxe, but it may be changed) and nearest neighbour rotation (the common rotation algorithm) and a gui to rotate current layer.
Tests are needed! If this is good, i can continue improving it.

Also, a scale3X function was added, in order to make other algorithms in the future.


nikhilCad:
Is it related to #42 ?

azagaya:
related, but i didn't implement rotsprite (yet).

PR #148, opened 2020-02-09
Better Outline Effect

AlphinAlbukhari:
I changed the outline algorithm.
Maybe the "Diagonal" option will no longer be useful.

Old:
old

New:
new


OverloadedOrama:
Ah, the code is much cleaner! It seems to be working better too, great job! But I think it should probably only replace the code that handles diagonal outlines, in order to leave the freedom of the user to choose what kind of outline they want to generate.

AlphinAlbukhari:
Please don't merge yet, it looks like something's wrong. I'll fix it

PR #154, opened 2020-02-15
Add option to hide the animation timeline

YeldhamDev:
If you're not doing any animations and just working on a isolated sprite, having it laying around is pointlessly consuming space. So here's a option to hide it.


OverloadedOrama:
This is a very nice idea, it will be very useful for single sprites indeed. Thanks!

PR #155, opened 2020-02-15
Make labels with tooltips have their mouse filtering set to "pass"

YeldhamDev:
Otherwise, their tooltips won't appear.


Calinou:
I wonder if we could throw a warning message at run-time on the engine level. We have a node configuration warning, but people may not always notice it, especially if creating a Label by code.

I suppose this would be problematic if you change the mouse filter in an animation though.

OverloadedOrama:
I did not know this to be honest. I put the tooltips, saw they weren't working and I gave up. This is good to know though, thanks!

PR #156, opened 2020-02-15
Hide presets in ColorPickers

YeldhamDev:
As their presets aren't saved, and for the canvas itself there's the palette system for that, they being visible is mostly pointless and creates noise.

Note that the "Add" button will still be visible due to a bug in Godot itself. It will be fixed with https://github.com/godotengine/godot/pull/36235 in a later release.


OverloadedOrama:
The presets for the left & right colors are actually being saved in .pxo files, but I did this before palettes were properly added, so I don't think it serves any purpose now and I'll remove it in v0.7.
Thanks!

PR #158, opened 2020-02-16
Make opening a project set the save path

YeldhamDev:
It's really annoying having to set the save path even after opening an existing project, so opening one now sets the save and export paths automatically.


OverloadedOrama:
Thanks!

PR #165, opened 2020-02-22
Ask for Confirmation when Pixelorama is about to quit and the Sprite is unsaved

Schweini07:


OverloadedOrama:
Looks good, thank you!

PR #168, opened 2020-03-02
Add palette import from a PNG file

novhack:
Adds ability to import a palette from a PNG file. Some palettes are distributed as PNG files with each color spanning exactly one pixel. Or it can be used to "scan" any image and get it's color palette.


OverloadedOrama:
Looks good, thank you!

PR #172, opened 2020-03-13
Fix LineEdit holding focus when it's not required

### LINKED ISSUE #167: Can't undo or redo changes after changing onion frame settings
After changing the values for "Past Frames" and "Future Frames", their ui elements seemed to keep focus, and I was unable to use the ctrl-z or ctrl-y shortcut in the main window.

Gaarco:
Fix #167


OverloadedOrama:
Thank you!

Issue #178, opened 2020-03-31
Feature: Improved New Canvas Dialog box


luiq54:
![image](https://user-images.githubusercontent.com/51489635/78023666-ae8b8080-7374-11ea-9c7d-051dd7f6da2a.png)
Features -
1) Lock canvas aspect ratio button 
2) Some default templates like 16x16, 32x32, 64x64, 128x128
3) It should remember the last created canvas size.

OverloadedOrama:
I'm adding the up for discussion label for the 3rd feature. The new image dialog gets the size from the default size, which can be changed in Preferences > Image. Do we want that, or do we want it to remember the last created canvas size?

I was thinking that, when the program first starts, it could get the size from the default size, and if the user changes size again, it could remember the last created size.

Calinou:
It seems GIMP remembers the last created size, so Pixelorama could do that as well for consistency.

luiq54:
Maybe let it remember the last created size only during the runtime of the app. On next launch it reverts back to the defaults.

PR #181, opened 2020-04-02
Improved CreateNewImage dialog #178

Gaarco:
As suggested in #178 I implemented some common templates and a checkbox to lock the aspect ratio, I think both can be useful and improve a bit the usability.
I wasn't interested to implement the "remember last used size" because there is already a field to define a default image size in the preferences and I thought it was a bit redundant and unnecessary.
I tested as far as i could and i didn't find any problems, hope you wont find any :grinning:


OverloadedOrama:
Didn't find any problems either. Thank you!

PR #182, opened 2020-04-03
Rotsprite-like Rotation

luiq54:
It upscales the image, rotates it and then downscales it.


OverloadedOrama:
Looks good, thanks!

PR #190, opened 2020-04-06
Add GIF export for Windows and Linux

novhack:
I tested it for some time but I would suggest not to merge until it's clear that everything works correctly on all supported platforms.

Supported platforms for now are Windows and Linux, both 32 and 64 bit. Some testing would be appreciated on 32 bit since all OS's I have are 64 bit.

To clone the plugin submodule: git submodule update --init
Plugin contains compiled GDNative libraries so no compilation should be needed.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "addons/godot-gifexporter"]
+	path = addons/godot-gifexporter
+	url = https://github.com/novhack/godot-gifexporter.git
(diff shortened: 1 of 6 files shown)


OverloadedOrama:
Looks good. Only issue I found was that if you change a layer's opacity and then export the image, the pixels of the canvas image itself get changed too, so if you want to bump it back to 100% opacity it doesn't work. That's unrelated to this PR however as it happens for png images too. It's an issue in ExportDialog.gd, in the blend_layers() method. I'll commit a fix for this right after I merge the PR.

Thank you!

Issue #175, opened 2020-03-24
Request: Custom Shortcut Key on Tools


krisnaxda:
i think shortcut key for pencil(O), and color picker(P) is too far and not efficient for left arm, it would be good if shortcut key can be changed. also how do i contribute to pixelorama for translation?

novhack:
You can contribute to translations here: https://crowdin.com/project/pixelorama

OverloadedOrama:
In a future version a way to re-map shortcut keys will be added... Maybe in v0.8, but I'm not sure yet.

PR #191, opened 2020-04-07
Add tools shortcuts binding to preferences

novhack:
Adds support for changing of shortcuts for main tools. Custom preset stores changes to `cache.ini` file so they get reloaded when Pixelorama starts. Default preset cannot be changed.


novhack:
I just realized that this solves #175.

OverloadedOrama:
Looks awesome, thank you!

PR #194, opened 2020-04-09
Move open/save functions to a separate file and clean up. Add ZSTD compression.

novhack:
I compared all compression algorithms that Godot offers (and included no compression results for comparison). For testing I created a file with 10 layers, 10 frames, each 1024x1024 pixels.

Save/Open times are in seconds measured on an NVMe drive. These times could be up to 10 times slower on a spinning hard drive but I hope that nobody will seriously try to create 1024x1024 pixel art...
||SAVE|OPEN|SIZE|
|---|---|---|---|
|RAW|0.5|0.7|400MB|
|DEFLATE|5.4|2.8|2.9MB|
|FASTLZ|2.7|2.2|5.8MB|
|GZIP|5.9|3|4.1MB|
|ZSTD| 2.5|1.9|2.2MB|

ZSTD came out on top with the best times and the best size reduction of *mere* **99.45%**. All hail our Facebook overlords!

### DIFF
--- a/Scripts/Dialogs/CreateNewImage.gd
+++ b/Scripts/Dialogs/CreateNewImage.gd
@@ -30,7 +30,7 @@ func _on_CreateNewImage_confirmed() -> void:
 	var width : int = width_value.value
 	var height : int = height_value.value
 	var fill_color : Color = fill_color_node.color
-	Global.control.clear_canvases()
+	Global.clear_canvases()
 	Global.layers.clear()
 	# Store [Layer name (0), Layer visibility boolean (1), Layer lock boolean (2), Frame container (3),
 	# will new frames be linked boolean (4), Array of linked frames (5)]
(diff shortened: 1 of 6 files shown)


novhack:
I also compared a RAW vs ZSTD on a more realistic scenario with only 10 frames 64x64. I wanted to be sure that ZSTD doesn't add some relatively big headers but it doesn't seem that it does. 

Save/Open times are negligible.
|RAW|ZSTD|
|---|---|
|160.5Kib|1.4KiB|

OverloadedOrama:
Looks good. On ImportSprites.gd, line 35, `Global.control.clear_canvases()` needs to be changed to `Global.clear_canvases()`. After this, it should be ready to get merged!

OverloadedOrama:
Thank you!

PR #195, opened 2020-04-10
Update the Godot theme to look more like godot's 3.x blue Theme

MatheusPese:
Fixed my last PR and made a new theme that looks like godot theme


OverloadedOrama:
Seems like the template resolutions got committed for this PR too. Maybe you could try having creating different branches for every PR you want to make.

We did see the new Godot theme though, and we like it, and the template code looks good too, so I'm going to merge this PR. Thank you!

PR #196, opened 2020-04-10
Added the left and right tool icon option to be saved and restored on  startup

MatheusPese:
Updated PreferencesDialog.tscn and PreferencesDialog.gd


OverloadedOrama:
Looks good, thanks!

PR #197, opened 2020-04-11
XDG basedir standard on X11 Godot platforms

sapient-cogbag:
Make Palettes/Brushes get loaded/saved in appropriate locations as specified by the [XDG basedir standard](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), for easier usage of standard linux/bsd packaging methods and for better per-user usability.


sapient-cogbag:
I do believe i have successfully got the Palette working nyaa <3

Calinou:
Why not use the Project Settings **Use Custom User Dir** property? It should handle everything for you.

sapient-cogbag:
The specific reason for me doing this is so the system handles $XDG_DATA_HOME and $XDG_DATA_DIRS correctly. To my understanding, the Godot setting does not handle them (especially the second one, since it cannot possibly know how to correctly do prioritised/overridable file loading) nya. The first one, again to my understanding, hardcodes it rather than checking the environment variable (though it might not). But one of the key things with this PR is that it means you can store read-only system files in /usr/share or wherever the system configured $XDG_DATA_DIRS, which is important for most packages nyaa.

Calinou:
I still think this is way too much code for something that is best left to the engine :slightly_frowning_face:

sapient-cogbag:
It just doesn't fit with the linux standard now, and as far as i know the engine doesn't handle it properly according to XDG basedir standards () so we just have to do it ourselves ;-;

sapient-cogbag:
(and yes, i am making a PKGBUILD :p)
(this is partially to allow native packages)

sapient-cogbag:
Both the Palette and Brushes should now functionally work with xdg basedir standard, with appropriate overriding behaviour ^.^

OverloadedOrama:
I've exported it on Windows so far and it works okay. Only problem is that it doesn't work when I run it inside Godot. It should create a pixelorama folder with the Brushes and Palettes folders inside of it, but it doesn't. It does on export though, which is the important part I guess.

The problem in the editor can be fixed if we create a "pixelorama" folder in the repo, and move "Brushes" and "Palettes" there. Could you do that? I think if you do it should be ready to get merged.

sapient-cogbag:
Will do!

sapient-cogbag:
It might be a good idea for me to make it so that when the data directories are empty it adds "res://" to it so it can load from internal data. I don't know if Brushes and Palettes actually get stored in the executable though. nya

OverloadedOrama:
Brushes and Palettes don't get stored in the executable, unfortunately. Their existence on the repo is purely to be there for the people that clone it. When I'm exporting the binaries, I'm copying the brushes and palette files manually.

sapient-cogbag:
Cool ^.^
this should be mergable then right :)
nya

OverloadedOrama:
Tested it on Ubuntu too. Works as expected. It seems that we can no longer include the Brushes and Palettes into the .zip itself like we used to, since it's loading/saving them elsewhere, so loading them from "res://" is an one-way road now. Probably copying them from res:// to the actual directories, if the directories don't exist. Which is all right, it'll save me some time from putting the files manually myself on every binary I export.

I'll merge it now, and we can figure out loading them from internal data some other time. Thank you!

OverloadedOrama:
Now that Godot 3.4 has new OS data directory getter methods (https://github.com/godotengine/godot/pull/49732), I wonder if the XDGDataPaths scripts can be simplified, or if it's needed anymore.

Calinou:
> Now that Godot 3.4 has new OS data directory getter methods ([godotengine/godot#49732](https://github.com/godotengine/godot/pull/49732)), I wonder if the XDGDataPaths scripts can be simplified, or if it's needed anymore.

I think the XDG script is no longer needed in 3.4 and later.

PR #198, opened 2020-04-13
Made arrow keys (with shift and control for speedup) work ^.^ nyaaa

sapient-cogbag:
Basically the title.
This patch means you can navigate the canvas area using the arrow keys with shift/control for moving them around ^.^
nya

### DIFF
--- a/Scripts/CameraMovement.gd
+++ b/Scripts/CameraMovement.gd
@@ -12,11 +12,87 @@ func _ready() -> void:
 	tween = Tween.new()
 	add_child(tween)
 
+
+# Get the speed multiplier for when you've pressed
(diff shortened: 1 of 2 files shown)


OverloadedOrama:
The only small issue with this is that Ctrl+Left and Ctrl+Right are used to go to the previous/next frame. Is it possible to avoid moving the canvas if these combinations are being pressed? Shift+Ctrl+Arrow isn't an issue, as it doesn't change the frame.

sapient-cogbag:
That should be trivial. Just an added condition nyaa

OverloadedOrama:
Awesome, thank you!

Issue #188, opened 2020-04-06
Editing all colors in a selection.


TheGoklayeh:
Modifying all the colors of a selection would be amazing
As seen in these two example, it is quite the poweful tool:
https://youtu.be/wiQJqddh4f8?t=761  
https://youtu.be/wiQJqddh4f8?t=1525

OverloadedOrama:
This is added in #199, so I'll go ahead and close this issue. Feel free to re-open it if you had something else in mind.

PR #199, opened 2020-04-13
Added HSV Adjust Dialog

luiq54:
It works for selection too!
Tests Needed!


OverloadedOrama:
This closes #188 right?

Also, could you rebase your branch so it won't be behind Orama-Interactive:master? There are conflicts in Main.tscn

luiq54:
@OverloadedOrama I have rebased it and run some tests to make sure it works! 😄

OverloadedOrama:
It seems to be a bit buggy if I press Cancel, and then redo. The idea of undoing on Cancel is a bit problematic. I think the best way to do it would be with a preview panel, like rotation has. If cancel is pressed, nothing happens. If OK is pressed, the changes happen. I plan to put previews on all other Image effects, probably for v0.8.

I've also noticed problems with the sliders. If I draw something black and then make Value go to 99, the resulting color is gray. If I drag the Value all the way to 0 and then back to 99, the resulting color is white. Shouldn't the same values result in the same results?

luiq54:
Yep, Preview panel seems the better option. This wil actually solve the second problem as well. Will add those today or tomorrow!

luiq54:
@OverloadedOrama I have added the previews and fixed the bugs as well. Also, the undo/redo functionality has been modified. Check it out and tell me if you encounter any more bugs! 😄

P.S. Dont merge it yet.  I am planning to add another commit regarding changes mentioned in contributing guidelines.

Scripts/Canvas.gd
```diff
@@ -1010,3 +1010,85 @@ func blend_rect(bg : Image, brush : Image, src_rect : Rect2, dst : Vector2) -> v
...
+						c.s = sat
+					elif(!selection_only):
+						var c: Color = layer.get_pixel(i,j)
+						var sat = c.s
+						if delta > 0:
+							sat = range_lerp(delta,0,100,c.s,1)
+						elif delta < 0:
+							sat = range_lerp(delta,-100,0,0,c.s)
+						c.s = sat
+						print(sat)
```
OverloadedOrama, line 1066: This is causing a lot of lag, make sure it remove it before the PR gets merged. Printing values multiple times in a frame generally causes a lot of lag, so make sure to only use it for specific debugging purposes and remove them before commiting! 😄

luiq54: oops,  my bad! 😅 . Yep I will fix it !!

Scripts/Canvas.gd
```diff
@@ -1010,3 +1010,85 @@ func blend_rect(bg : Image, brush : Image, src_rect : Rect2, dst : Vector2) -> v
...
+				for j in range(layer.get_height()):
+					if(selection_only and Global.selected_pixels.has(Vector2(i,j))):
+						var c: Color = layer.get_pixel(i,j)
+						var val = c.v
+						if delta > 0:
+							val = range_lerp(delta,0,100,c.v,1)
+						elif delta < 0:
+							val = range_lerp(delta,-100,0,0,c.v)
+						
+						c.s = val
```
OverloadedOrama, line 1081: Small typo here, this should be `c.v`, not `c.s`.

luiq54:
Fixed the bugs and went through the code once! I have added Translations Strings. I think this is ready to merge.

OverloadedOrama:
Yeah I think so too. Thank you!

PR #202, opened 2020-04-15
Add a Delete Button and a Copy Button for frames

Schweini07:
You can now click on the delete button and you do not have to open the popup menu


novhack:
You can also delete with a middle mouse click on a frame.

Schweini07:
Quick question then: Are there other shortcuts for the actions in the popupmenu? Because copying frames for instance, is also a bit cumbersome. Thanks!

Schweini07:
Now there's also a copy button so that the PR is more useful

Schweini07:
Just noticed, I forgot the tooltips, please wait a bit

Schweini07:
Should work now, I hope I didn't conflict with the HSVDialog PR

Schweini07:
I did conflict, but should be resolved now

OverloadedOrama:
Using the latest commit, I only see the remove button and its hint tooltip is the same as the add button. Did you put the copy button on the scene, or only as a file?
image

Schweini07:
Should work now, I forgot the AnimationTimeline.tscn

OverloadedOrama:
Looks good, I'll go ahead and merge. Thanks!

PR #210, opened 2020-04-18
Improve the checker board of transparent background

Kinwailo:
image

I added option to preferences under "Grid&Guides", but I changed the title to "Canvas". I cannot set the ui fill the cell, each time run or save, Godot resize the ui.


OverloadedOrama:
Amazing! Just what Pixelorama needed, thank you!

PR #215, opened 2020-04-21
Open last edited project at startup or at will feature

jegor377:
Created feature that will alllow user to open last edited project on Pixelorama startup or open last edited project at will.


OverloadedOrama:
Looks good, thank you!

PR #216, opened 2020-04-21
Translation completion for new feature.

jegor377:
Completed the hole in translation for the new open last project on startup or at will feature.


OverloadedOrama:
Thanks!

PR #217, opened 2020-04-22
Export multiple files in multiple directories feature

jegor377:
New feature that will allow user to export animation frames in multiple filed in multiple directories depending on frame tag.
Each animation frame file will have id starting from 1 for each animation from now.


OverloadedOrama:
From what I understand, the only difference between "All frames as multiple files" and "All frames as multiple files in multiple directories" is when the animation has tags, right? I really like the idea, but, wouldn't it be better if we only kept "All frames as multiple files" as an option, and add a checkbox so the users can choose if they want to export the tagged frames in unique directories? I think the way it's currently being done is a little confusing for users.

jegor377:
No problem. I can do this so that user will just choose "All frames as multiple files" and then just a little checkbox will popup asking if user want files to be stored in different directories named after animation tags.

OverloadedOrama:
Okay, I think that sounds good!

jegor377:
Okay I fixed this but I'm not sure if I commited it correctly. :/

OverloadedOrama:
Looks good, thank you!

PR #221, opened 2020-04-25
Add autosave feature with backup of unsaved new projects

novhack:
Pixelorama will now automatically save based on user preferences.
If user created a new image and haven't manually saved it a backup save file will be created in user:// on autosave. If Pixelorama crashes and backup file path is present in config file user will have an option to restore it or not.
Screenshot_20200425_224148

I also moved "Open last project on startup" down to autosave options and renamed "OK" button to "Close" so it won't evoke that preferences are set when "OK" is pressed but immediately.
Screenshot_20200425_223622


Scripts/Main.gd
```diff
@@ -189,10 +189,19 @@ func _ready() -> void:
...
+		# Temporatily stop autosave until user confirms backup
+		OpenSave.autosave_timer.stop()
+		$BackupConfirmation.dialog_text = $BackupConfirmation.dialog_text % OS.get_user_data_dir()
+		$BackupConfirmation.popup_centered()
+	else:
+		if not Global.config_cache.has_section_key("preferences", "open_last_project"):
+			Global.config_cache.set_value("preferences", "open_last_project", true)
+		if Global.config_cache.get_value("preferences", "open_last_project"):
+			Global.open_last_project = Global.config_cache.get_value("preferences", "open_last_project")
 		load_last_project()
```
OverloadedOrama, line 205: Should this line be intended one more time? The last project is being loaded even if the "open last project" setting is turned off in the Preferences.

novhack: Yes. Indentation on line 205 was wrong.

novhack:
So I did some wanted changes:
- Autosave is no longer performed on files that were created manually by user. Instead a backup file in user:// is created for every autosaved file. When user manually saves this backup is deleted.
- If user closes Pixelorama manually (either saves or decides not to save) the backup is deleted.
- If there are multiple instances of Pixelorama it should no longer be an issue. Example: If two Pixelorama instances crash and leave backups - on the next startup the first backup found is offered for reload. If that backup is deleted by manual save the next startup will offer the second backup for reload. When multi-project support is added to Pixelorama a possibility to reload all backup at once should be added.

OverloadedOrama:
Looks great to me. Thank you!

Issue #192, opened 2020-04-08
Pixel Perfect Pencil


novhack:
Every major pixel art tool has this feature to enable fast drawing without doubles.
https://rickyhan.com/jekyll/update/2018/11/22/pixel-art-algorithm-pixel-perfect.html

Calinou:
Good catch. It seems this can be done just after the Bresenham line algorithm?

OverloadedOrama:
#222 implemented this, so I'll close the issue. Feel free to re-open it you think that something is missing/needs to change!

PR #222, opened 2020-04-26
Pixel Perfect mode for brush, eraser and lighten/darken

dragonfi:
Screenshot from 2020-04-26 12-18-36

Added a pixel perfect mode for the pencil and similar brush-like tools

See issue: https://github.com/Orama-Interactive/Pixelorama/issues/192

Unrelated question: I see that you have nice variable names and the code is quite readable. Still, I feel that some methods are too long and there is a lot of repetition. Would you accept refactor-only pull requests?

### DIFF
--- a/Main.tscn
+++ b/Main.tscn
@@ -637,6 +637,23 @@ value = 1.0
 allow_greater = true
 ticks_on_borders = true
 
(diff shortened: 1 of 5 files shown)


OverloadedOrama:
Looks good! As for your question, I've been thinking of refactoring the code, but I was too busy working on new features. However, because v0.7 is about to be released, I'm worried that refactoring will cause issues. I was planning to do some refactoring immediately after v0.7 is released.

So, yes, refactor-only pull requests are welcome and I would really appreciate them, but I think it's best that they happen after v0.7 is released.

Thank you!

PR #230, opened 2020-05-08
Added selective animation tag export for Spritesheets

Wishdream:
This lets people export **a single tag** or **all frames** in spritesheet export tab:

image

image

In the future, I want to work on this more to also carry this over to Animation as well and also the `Create new directory for each frame tag` over to the Spritesheet export.

### DIFF
--- a/Translations/Translations.pot
+++ b/Translations/Translations.pot
@@ -167,6 +167,12 @@ msgstr ""
 msgid "Frame"
 msgstr ""
 
+msgid "Frames:"
+msgstr ""
+
+msgid "All Frames"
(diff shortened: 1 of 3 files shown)


OverloadedOrama:
It looks good. I was also thinking that we could have an extra option for "All Frames", to categorize the frames based on their tag instead of Rows and Columns. So, for example, each row would contain frames of the same tag. That way, the user can export organized texture atlases. I'm not sure if that would be useful though, so let me know if you like this idea.

Anyway, I'll go ahead and merge this. Thank you!

PR #234, opened 2020-05-16
Update gitignore

aaronfranke:
I updated the gitignore to be based on other Godot gitignores. It's useful to keep this in sync even if not everything is directly relevant. For example, opening the project in a C#-enabled version of Godot will generate the `.mono/` folder even if the project doesn't use C# anywhere.

Also, I ran a formatting script, there was a single missing newline in a shader, so I added it.


OverloadedOrama:
Great, thank you!

PR #237, opened 2020-05-17
Update godot-gifexporter submodule.

novhack:
Fixes an error that was displaying when Pixelorama project was opened in the Godot Editor.
Sorry that it took so long to fix for such annoying little mistake.


OverloadedOrama:
Thank you!

PR #243, opened 2020-05-22
Refactoring image_menu_id_pressed method in Main.gd

jegor377:
I've moved the code from each "match" case into a seperate method to make it more readable.


OverloadedOrama:
Good idea! Thanks!

PR #249, opened 2020-05-30
Remove fallback for `OS.min_window_size` now that Godot 3.2 is out

Calinou:
The latest Pixelorama versions require Godot 3.2 to run.


OverloadedOrama:
Thanks!

PR #250, opened 2020-05-31
Gimp palette import fixes

rob-a-bolton:
Adds support for the newer variant of gpl files where "Name: " is used for palette name, and extracts palette name from comment if present (some converter/exporter must do this, as it was present in 4 of the palettes I had downloaded).
Also supports palettes where no name is supplied for the colours.  

These changes fix failed imports of palettes using this new format, and palettes with unnamed colours.  
Tested on a handful of GIMP palettes and appears to work fine.


OverloadedOrama:
Looks good, thank you!

PR #255, opened 2020-06-04
Move pixel perfect and mirror code to drawers class

Kinwailo:
The pixel perfect don't work correctly when brush size is not 1, I forced it disabled when brush size is not 1.


OverloadedOrama:
Looks good, thank you!

PR #282, opened 2020-07-13
Add overwrite option to pencil.

### LINKED ISSUE #278: Allow drawing with transparent color
I like to use only the pencil tool and the color picker. To erase something, instead of selecting the eraser tool, I pick up the transparent color and draw over it. 
In Pixelorama, I can pick up the transparent color, but nothing happens when I try to draw with it.

Kinwailo:
Overwrite color instead of blending it.

Fixes #278


OverloadedOrama:
Looks good! It was a good idea to make this behavior toggle-able. Thank you!

PR #300, opened 2020-08-09
Added TransparentChecker Background to Dialog Previews

### LINKED ISSUE #206: Include a background to all canvas previews
Currently previews in the export dialog and in the timeline frames have all transparent backgrounds so the drawings are with some themes hard to see.
I propose to add a checker board background to all previews mentioned (or more solid color for timeline frame previews).
Screenshot_20200417_005000
It should be considered if it's actually necessary to show a preview in the frame button since it mostly becomes unrecognizable (maybe a simple glyph that signifies that there is something in the layer would be better and more visible).
Screenshot_20200417_004949

luiq54:
Fixes #206 .
Used a modified version of Canvas' TransparentChecker Shader .

### DIFF
--- a/src/Shaders/TransparentCheckerDialog.shader
+++ b/src/Shaders/TransparentCheckerDialog.shader
@@ -1,0 +1,17 @@
+shader_type canvas_item;
+render_mode unshaded;
+
+uniform float size = 10.0;
+uniform vec4 color1 : hint_color = vec4(0.7, 0.7, 0.7, 1.0);
+uniform vec4 color2 : hint_color = vec4(1.0);
+
+void fragment() {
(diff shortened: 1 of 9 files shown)


luiq54:
There is some funky behavior for Invert and Desaturate Dialogs . Other dialogs have been checked for problems.

luiq54:
Now that I think about it, your solution would have been easier to implement. I will look into what is causing the issues (for HSV and Invert ) or else we use your solution.

luiq54:
I checked the code and I can't find what's causing this issue. I think your solution is the better one and should be implemented instead. So I will close this pull request.

PR #315, opened 2020-08-25
Add PAL palette format import

### LINKED ISSUE #307: .pal file type for color palettes.
Title says it all. .pal file should be also supported since its one of the major color palette file format as well as all other drawing software also supports it .

matthewpaul-us:
Added a PAL importer to fix #307. I used a palette from Lowspec to test: [Isa's Limited Ramps](https://lospec.com/palette-list/isas-limited-ramps)

**Before palette import**
Before I imported the palette, you can see that it doesn't exist.
image

**Adding the accepted file formats**
image

**Post-import Palette**
The palette takes the file name as the palette name. Note there are 28 colors which match the number of colors in the palette. 
image

Let me know if I need to change anything.

### DIFF
--- a/src/Autoload/HTML5FileExchange.gd
+++ b/src/Autoload/HTML5FileExchange.gd
@@ -47,7 +47,7 @@ func _define_js() -> void:
 		canceled = true;
 		var input = document.createElement('INPUT');
 		input.setAttribute("type", "file");
-		input.setAttribute("accept", "application/json, .gpl, image/png, image/jpeg, image/webp");
+		input.setAttribute("accept", "application/json, .gpl, .pal, image/png, image/jpeg, image/webp");
 		input.click();
 		input.addEventListener('change', event => {
 			if (event.target.files.length > 0){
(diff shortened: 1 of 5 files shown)


OverloadedOrama (commented):
I think it looks good! Just a minor detail, I think it's best to change the PaletteImportFileDialog's filters in its own scene (PaletteImportFileDialog.tscn), instead of PalettePanelContainer.tscn.

It would also be good to make it possible to import PAL palette files from the Web version too, by making changes to the `HTML5FileExchange.gd` script. It's not a big issue though, I can do that myself if you want.

matthewpaul-us:
Ah! I missed it on the file dialog itself. I'll make that change definitely. Thanks for the catch!

I'll also look into the HTML5FileExchange to see the difficulty. If it's not too bad, I'll include it.

matthewpaul-us:
I moved the selection filter into the file dialog.

I attempted the HTML change, but I wasn't sure how to test it. I exported it to HTML5 and then ran it using a python web server. When I went to import the palette, it threw a javascript error. I think it's because I am using the same palette_data as the gpl, when I need the text back from the javascript. I'll keep looking at it.

image

matthewpaul-us:
I was able to get back to this and saw my issue. I had forgotten the s on ends_with which caused it to fail. After fixing this and an issue with javascript comparison not liking the not-equals, it works in the web as well.

OverloadedOrama (approved):
Okay I think everything is good now. Thank you!

PR #318, opened 2020-08-26
Add transparent checker to export dialog and animation cels

### LINKED ISSUE #206: Include a background to all canvas previews

matthewpaul-us:
I added transparent backgrounds to both the export dialog and the animation cels. Fixes #206 if there's nothing holding this issue up.

**Export Dialog**
I had to make a few changes to the node structure on the dialog, as the scroll container didn't take to kindly to me adding a child checker. Neither did the grid container. I ended up putting them in a panel and sticking the checker behind them.
image

Here is a preview of the new node structure.
image

**Animation Cels**
These were as simple as adding the child and selecting display behind parent.
image

### DIFF
--- a/src/UI/Dialogs/ExportDialog.gd
+++ b/src/UI/Dialogs/ExportDialog.gd
@@ -15,7 +15,7 @@ onready var export_progress_popup = $Popups/ExportProgressBar
 onready var export_progress_bar = $Popups/ExportProgressBar/MarginContainer/ProgressBar
 
 onready var animation_options_multiple_animations_directories = $VBoxContainer/AnimationOptions/MultipleAnimationsDirectories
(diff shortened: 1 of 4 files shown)


OverloadedOrama (approved):
Looks great, thank you!

PR #327, opened 2020-09-05
Fix for remove_backup_by_path.

jegor377:
The method tried to remove backups section project_path key but it didn't exist so Godot reported error. Now it doesn't because it first checks if the project_path exists in backups section.


OverloadedOrama:
Superseded by 497f0e03cb0f061f9b1379fdf846ad66f2dbbd2d. Thanks!

PR #342, opened 2020-09-28
Theme edits

Erevoid:
Changed all the themes to be more consistent, buttons and panel containers have rounded edges. Note that the spacings have been switched too, so it needs subtle remakes in the placing of each element. Especially the tool buttons. Also check the themes on preferences (need better spacing) and rulers are not included.


OverloadedOrama (approved):
Looks good, thanks!

Issue #306, opened 2020-08-11
Timeline Improvements


OverloadedOrama:
**Feature/enhancement description:**
v0.7 came with a new timeline, but I think there is still a lot of room for improvement. Some ideas that I have are:

- [x] Frame-specific delay.  (Issue #267 - implemented in #357)
- [x] Drag and drop individual cels to move them around. (27852b2694537287556e1e943428868e5572e4c4)
- [x] Hold <kbd>Control</kbd> and mouse wheel to resize cels in the timeline. (7f1594e1bc890dc29f8dad59f1aa2c0a23454974)
- [x] Create a button for each for frame instead of a label. Left clicking it selects the frame, right clicking it opens a menu with options suck as remove, clone, move & properties. (39b0baca84c0f4e27a3f593255edaf9085be2002 and ecd08ce132355c07d261b1867085dc09d0ea13fc)
- [x] Change the right click menu of the cels. It's currently frame-related, while it should be cel-related. (ecd08ce132355c07d261b1867085dc09d0ea13fc)
- [x] Clicking on a layer button selects the layer. Double clicking it allows renaming. (e7a5b4553611483a51fc8eb5d1902a5cd55040c2)
- [x] Put Delete and Link/Unlink in the cel right click menu. (1a87321437e9dc8da4b2415bbe71a39d02fad01c)
- [x] Drag and drop entire frames and layers to move them around. Their cels will also get moved with them. (e9bf310934d2bbc0cdb2b35dfba6af00e530f366)
- [x] Press <kbd>Control</kbd> to select multiple cels, and <kbd>Shift</kbd> to select multiple cels along with the cels between them. Pressing <kbd>Control</kbd> a on a selected cel would de-select it, unless it's the only selected cel. (de850ce8a11299f263fb867de7ecc3c792783fc3)
- [x] Changes made to a selected cel will also affect all other selected cels. That includes tools (like drawing with pencil) and image effects. (de850ce8a11299f263fb867de7ecc3c792783fc3)
- [ ] Figure out where to put cel opacity and layer opacity. Layer opacity does not exist yet.

The list may be edited to include more ideas in the future.

novhack:
I have a big monitor and I would really like an option to resize cels. :slightly_smiling_face:

Erevoid:
Got some more ideas regarding **tags**:

- The ability to move the tag with its cels left or right in the timeline (in regards to other tags or single untagged frames)
[...]

PR #344, opened 2020-10-02
Add buttons for moving the current frame left or right

Schweini07:
I finally should have everything.
aufnahme-2020-10-02-145848


OverloadedOrama (approved):
I think it looks good, and it's a good step towards #306. Thanks!

Issue #284, opened 2020-07-25
Add cut and paste


yonderbread:
There's currently no cut and paste functionality when selecting a region. Could you add this?

brendonferreira:
+1

OverloadedOrama:
Paste is already there, and cut just got implemented in #345!

PR #345, opened 2020-10-08
Cut option

PinyaColada:
I made cut-option, it appears in the top bar the function and you can call the cut with c-x


OverloadedOrama (approved):
Looks good, thank you! This also closes #284.

PR #346, opened 2020-10-09
Sync

PinyaColada:
There's a bug that if you use cut, paste and delete while moving the selection rectangle this goes to the position it started moving so with this few lines it is fixed


OverloadedOrama (approved):
Looks good, thanks!

PR #357, opened 2020-10-17
Frame properties

### LINKED ISSUE #267: Can each frame have a custom delay?
Geia sou pali :)

I noticed PO uses FPS for the animation. Is there a way to change a custom delay per frame?
PO can export to GIF. Can it also import a GIF and show the frames?

Thanks

PinyaColada:
Frame properties update:
- Now at right click a cell it will popup and option of "frame properties" and entering in that popup it will open a windows, in that windows it will display the frame of the cell you clicked and also how many times that frame should last (for example 1 x means 1 time which is the standart but 6 x means it will last 6 times the time it should last, at 6 FPS per sencond that is a second complete)
- The serialize() and deserialize() function has been changed to add the list frame_duration for being able to save/unsave
- The gif exporter works with the frame_duration
- The change of frame time it register as an action so it can be undo/redo
- The translation.pot has been modified

This is my first update a little big and my first time working in an open source project, so I'll be glad to hear any criticism but hope you like it :)

**Overloaded Edit: This closes #267 and partially addresses #306.**

### DIFF
--- a/src/UI/Timeline/CelButton.gd
+++ b/src/UI/Timeline/CelButton.gd
@@ -7,6 +7,7 @@ onready var popup_menu : PopupMenu = $PopupMenu
 
 
 func _ready() -> void:
+	popup_menu.add_item("Frame properties")
 	hint_tooltip = tr("Frame: %s, Layer: %s") % [frame + 1, layer]
 	if Global.current_project.frames[frame] in Global.current_project.layers[layer].linked_cels:
 		get_node("LinkedIndicator").visible = true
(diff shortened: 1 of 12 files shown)


src/UI/Timeline/CelButton.gd
```diff
@@ -7,6 +7,7 @@ onready var popup_menu : PopupMenu = $PopupMenu
 
 
 func _ready() -> void:
+	popup_menu.add_item("Frame properties")
```
OverloadedOrama, line 10: It would be better if this was in `CelButton.tscn`, along with the rest of the `PopupMenu` node items. When clicking the node, you will see an "Items" option on the top bar. It would be better if this gets added there, just for the sake of keeping grouped stuff together.

OverloadedOrama (commented):
Looks good, amazing job! The only thing I noticed that is not working properly is the canvas preview animation. It's working properly when the play button on the timeline is pressed, but not when the smaller play button next to the mini canvas is pressed. I think that would require some changes to `CanvasPreview.gd`. Let me know if you wanna give it a shot, otherwise I can try and fix the issue myself.

Other than that, impressive job!

src/UI/Timeline/CelButton.gd
PinyaColada: - About the CanvasPreview.gd: that's true I didn't see that play button and I'll try to do it myself
- About the CelButton.gd that add_item it was a provisional thing but I end it up forgeting about it I'll fixit
- Another thing now that I'm thinking about that with the new serialize and deserialize it could be problems opening old .pxo so I should fix that or there is something for this cases?

OverloadedOrama: Opening old .pxo files should work, you already put a check to see if the loaded data have `frame_duration` (line 337 in Project.gd). I tested opening an old pxo file and it worked, so I think it's ready.

PinyaColada: Okey! I've made the changes hope everything is ok!

OverloadedOrama (approved):
Looks good, thank you!

Issue #132, opened 2020-01-15
Brush improvements


Erevoid:
Here are some ideas for improving the brushes:

Brush tilting: The ability to rotate the brushes

Middle click for project brush removal: With its current format, it's quite easy to accidentally erase the project brush you already have, so the ability of being able to remove the project brushes with **only** the middle click would solve that problem.

The ability to save your project brushes inside the file brushes from within the program: That could be potentially done by right clicking on the brush and a text appearing saying: save to file brushes. one step further would be the ability to add random brushes that way too. Similarly you could right click on a project brush and select: Save as file brush. A popup window can then open, enabling you to name your brush (if it's a random brush the % will be added automatically) and if you add another brush with the same name as a previous one, then it becomes a random brush with those two brushes (or more if it's done more than two times)

OverloadedOrama:
More ideas:

- [ ] Load .gd files (GDScript files) as digital brushes. Digital brushes can be a new type of brush, that uses logic to handle drawing, scaling, and potentially other parameters that can be exposed as settings to the user. This idea is still under consideration.
- [ ] Add a refresh button that reloads brushes, like GIMP has.
- [ ] Make the UI more clear, so it's easier for the user to understand which type is each brush. A lot of people don't know about random brushes and they think of them as bugs at the beginning.
- [ ] Change how random brushes are being stored and loaded to solve #337 and to improve user experience.
- [ ] Easy way to save project brushes as file brushes. Could come hand in hand with the selection system improvements.

PR #362, opened 2020-10-21
Use '~' prefix instead of '%' for random brushes

### LINKED ISSUE #337: Brush filenames with '%' trigger rpmlint warning when packaging Pixelorama as RPM
**Pixelorama version:**
0.8, but earlier versions too

Mageia Cauldron (Linux)

**Issue description:**
I'm packaging Pixelorama for my Linux distro Mageia (using RPM), and the linter tool `rpmlint` generates warnings on the file paths using the `%` character, as it's normally used for macros in RPM spec files:
```
$ rpmlint RPMS/noarch/pixelorama-0.8-1.mga8.noarch.rpm 
pixelorama.noarch: W: unexpanded-macro /usr/share/pixelorama/Brushes/Grass/%Grass1.png %Grass1
pixelorama.noarch: W: unexpanded-macro /usr/share/pixelorama/Brushes/Grass/%Grass2.png %Grass2
pixelorama.noarch: W: unexpanded-macro /usr/share/pixelorama/Brushes/Grass/%Grass3.png %Grass3
pixelorama.noarch: W: unexpanded-macro /usr/share/pixelorama/Brushes/Grass/%Grass4.png %Grass4
pixelorama.noarch: W: unexpanded-macro /usr/share/pixelorama/Brushes/Grass/%Grass5.png %Grass5
pixelorama.noarch: W: unexpanded-macro /usr/share/pixelorama/Brushes/Star/%star1.png %star1
pixelorama.noarch: W: unexpanded-macro /usr/share/pixelorama/Brushes/Star/%star2.png %star2
pixelorama.noarch: W: unexpanded-macro /usr/share/pixelorama/Brushes/Star/%star3.png %star3
pixelorama.noarch: W: unexpanded-macro /usr/share/pixelorama/Brushes/Star/%star4.png %star4
pixelorama.noarch: W: unexpanded-macro /usr/share/pixelorama/Brushes/Star/%star5.png %star5
pixelorama.noarch: W: unexpanded-macro /usr/share/pixelorama/Brushes/Star/%star6.png %star6
```

Would it be an option to change their names to use another kind of prefix?

The warning is not actually an issue for the RPM package, but my distro treats `unexpanded-macro` warnings as errors, so it's bothersome (I'd have to patch pixelorama to change the names, or add a distro-wide exception for this warning on pixelorama).

**Steps to reproduce:**
- Build a RPM package for pixelorama, e.g. http://svnweb.mageia.org/packages/cauldron/pixelorama/current/SPECS/pixelorama.spec?view=markup
- Run `rpmlint` 1.11 (latest version) on the resulting package

akien-mga:
Fixes #337.

It's not necessarily the best fix but according to https://en.wikipedia.org/wiki/Filename#Reserved_characters_and_words `~` doesn't seem to be reserved, so it might be a safer bet.
It works for me on Linux but I haven't tried other OSes to confirm.

Note that a `~` *suffix* is used by some code editors to mark backup versions of changed files, which might be a source of confusion (but does show that `~` in filenames seems to be safe). It doesn't seem to conflict with `~` that expands to `$HOME`, at least in bash, as long as you don't use `~` alone or with `~/somepath`.


akien-mga:
Note also that this would break compatibility for users who currently rely on the `%` prefix, who would need to rename their brushes. This could be improved to keep support for both prefixes, or as mentioned in https://github.com/Orama-Interactive/Pixelorama/issues/132#issuecomment-709286138, the whole system could be changed to something a bit more advanced.

I'm completely fine with this not being merged as is, I use it for my distro package as a workaround but it's fine with me if another, longer term solution is worked on for #337.

OverloadedOrama (approved):
Thanks for taking the time to fix this. We will most likely change how the brush system works, although I'm not sure yet how it will be and when it will happen, but for now this PR will do!

PR #368, opened 2020-10-25
Add feature request #358

### LINKED ISSUE #358: About the current Tile Mode
Hey guys I'm back!  I've been making tilemaps in Pixelorama and I tried using the Tilemode for the texture and base appearance of  my tilemap.  I created a new tab cause I need them seperated, and when I'm done pixeling using tile mode I couldn't find where the real one is because of the copies that surrounds it looks the same and have the same transparency. I just hope there is an indicator when using tile mode that will indicate where the real one is.

Here is my Suggestion:
       make the other copies that surrounds it (when you use tile mode) translucent like what Piskel did.  The other copies that surrounds it will have lesser transparency so the real one sticks out among all. 

Thats all folks thank you!

dasimonde:
You can change the opacity of the surrounded tiles in tile mode.
You can change the opacity in the preference dialog under the subitem Canvas/transparency

**Overloaded Edit:** Closes #358

### DIFF
--- a/src/UI/Canvas/TileMode.gd
+++ b/src/UI/Canvas/TileMode.gd
@@ -23,9 +23,7 @@ func _draw() -> void:
 		# Mostly used to hide the grid if it goes outside the canvas boundaries
 		draw_rect(Rect2(pos, size), Global.default_clear_color)
 
-	var tilemode_opacity = 1.0
-	if Global.config_cache.has_section_key("preferences", "tilemode_opacity"):
-		tilemode_opacity = 1.0 - Global.config_cache.get_value("preferences", "tilemode_opacity")
+	var tilemode_opacity = 1.0 - Global.tilemode_opacity
 
 	for i in range(Global.current_project.layers.size()):
 		var modulate_color := Color(1, 1, 1, current_cels[i].opacity - tilemode_opacity)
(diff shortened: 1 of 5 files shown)


src/UI/Canvas/TileMode.gd
```diff
@@ -23,8 +23,12 @@ func _draw() -> void:
 		# Mostly used to hide the grid if it goes outside the canvas boundaries
 		draw_rect(Rect2(pos, size), Global.default_clear_color)
 
+	var tilemode_opacity = 1.0
+	if Global.config_cache.has_section_key("preferences", "tilemode_opacity"):
+		tilemode_opacity = 1.0 - Global.config_cache.get_value("preferences", "tilemode_opacity")
```
OverloadedOrama, line 28: There is a small issue here. If the user tries the tile mode without changing its opacity from the preferences, `tilemode_opacity` remains 1.0, so `modulate_color`'s alpha channel is 0, and tile mode is invisible. Perhaps `tilemode_opacity`'s default value could be 0.0 instead, in line 26. Or we can try getting rid of the if statement and just do `var tilemode_opacity = 1.0 - Global.tilemode_opacity`. If I remember correctly, `Global.tilemode_opacity` should change based on what is saved in the config cache anyway.

dasimonde: The issue should be fixed now. I removed the if statement. The default value for the opacity at the beginning is 1.0, now.

OverloadedOrama (approved):
Looks good, thank you!

PR #376, opened 2020-11-06
Implement shrink option

### LINKED ISSUE #140: Feature request: support HiDPI display / editor display scale setting
Hi, thanks so much for the amazing work with Pixelorama!

It would be brilliant if there could be support for HiDPI displays as at the moment the interface is quite small in scale and a bit hard on the eyes:

pixelorama

Godot has a "Display Scale" setting for the editor that would solve the problem beautifully if it were possible to bring it across:

godotscaling

Thanks for considering, and for this great project. :)

gschwind:
Hello,

On my computer Pixelorama was unusable because of tiny size of the GUI. I using Linux with a screen with HiDPI but godot does not support HiDPI for Linux.

This patch some how fix the issue with an option to set the shrink value. This values is used by godot to scale the whole application. The option is available in the configuration panel of the godot project and it can be changed in live by the application. The purpose of this patch the add an option in the Parameter>Startup section to do the job.

I see few possible issue, in particular if the value is set at a too high value and the parameter screen is not usable anymore. Otherwise it work nice.

Let me know if you want some change.

Best regards.

Overloaded edit: Closes #140

### DIFF
--- a/src/Main.gd
+++ b/src/Main.gd
@@ -63,6 +63,9 @@ func setup_application_window_size() -> void:
 		return
 	# Set a minimum window size to prevent UI elements from collapsing on each other.
 	OS.min_window_size = Vector2(1024, 576)
+	
+	get_tree().set_screen_stretch(SceneTree.STRETCH_MODE_DISABLED,
+		SceneTree.STRETCH_ASPECT_IGNORE, Vector2(1014,576), Global.shrink)
 
 	# Restore the window position/size if values are present in the configuration cache
 	if Global.config_cache.has_section_key("window", "screen"):
--- a/src/Autoload/Global.gd
+++ b/src/Autoload/Global.gd
@@ -43,6 +43,7 @@ var default_clear_color := Color.gray
 # Preferences
 var pressure_sensitivity_mode = Pressure_Sensitivity.NONE
 var open_last_project := false
+var shrink := 1.0
 var smooth_zoom := true
 var theme_type : int = Theme_Types.DARK
 var default_image_width := 64
--- a/src/Preferences/PreferencesDialog.gd
+++ b/src/Preferences/PreferencesDialog.gd
@@ -3,6 +3,7 @@ extends AcceptDialog
 # Preferences table: [Prop name in Global, relative node path, value type, default value]
 var preferences = [
 	["open_last_project", "Startup/StartupContainer/OpenLastProject", "pressed", Global.open_last_project],
+	["shrink", "Startup/ShrinkContainer/ShrinkHSlider", "value", Global.shrink],
 	["smooth_zoom", "Canvas/ZoomOptions/SmoothZoom", "pressed", Global.smooth_zoom],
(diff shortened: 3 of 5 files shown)


OverloadedOrama (approved):
This is awesome! I've been struggling with this for so long, but I always had issues with the mouse coordinates on the canvas. This seems to be working fine and I like the implementation. Thank you!

Calinou:
> I using Linux with a screen with HiDPI but godot does not support HiDPI for Linux.

To be precise, the Godot editor supports hiDPI on all platforms, but exported projects can't expose proper hiDPI support until https://github.com/godotengine/godot/pull/21446 is merged.

src/Main.gd
```diff
@@ -63,6 +63,9 @@ func setup_application_window_size() -> void:
 		return
 	# Set a minimum window size to prevent UI elements from collapsing on each other.
 	OS.min_window_size = Vector2(1024, 576)
+	
+	get_tree().set_screen_stretch(SceneTree.STRETCH_MODE_DISABLED,
+		SceneTree.STRETCH_ASPECT_IGNORE, Vector2(1014,576), Global.shrink)
```
Calinou, line 68: Out of curiosity, why is this `1014` instead of `1024`?

Also, you can get this value dynamically using `ProjectSettings.get_setting("display/window/size/width") and `ProjectSettings.get_setting("display/window/size/height")`. This makes the code easier to reuse across projects.

gschwind: It is a typo :), I can fix it

Best regards

gschwind:
> > I using Linux with a screen with HiDPI but godot does not support HiDPI for Linux.
> 
> To clarify, the Godot editor supports hiDPI on all platforms, but exported projects can't expose proper hiDPI support until [godotengine/godot#21446](https://github.com/godotengine/godot/pull/21446) is merged.

Ok, I currently using godot 3.2.3-stable. Anyway I currently using the patch for 3x shrink while for my screen 2x should be the expected value, so it may be useful some how. I'm fine if it is not merged :)

Best regards

src/Main.gd
gschwind: About ProjectSettings.get_setting("display/window/size/width"), I just followed the line above where OS.min_window_size is set to Vector2(1024, 576). The 3rd parameters of set_screen_stretch is minsize.

gschwind:
Hello, I did improved patch, with comment, a fix to add location in the computation of current_pixel and the fix of typo.

gschwind@774f7dfe2da053206f220b89afca62bafc7b4103
gschwind@ed8e898a9f0e6c557ff25edfd1d5716d8256413b

OverloadedOrama:
> Hello, I did improved patch, with comment, a fix to add location in the computation of current_pixel and the fix of typo.
> 
> [gschwind/Pixelorama@774f7df](https://github.com/gschwind/Pixelorama/commit/774f7dfe2da053206f220b89afca62bafc7b4103)
> [gschwind/Pixelorama@ed8e898](https://github.com/gschwind/Pixelorama/commit/ed8e898a9f0e6c557ff25edfd1d5716d8256413b)

I included your changes in bd10680f636546800fdcbe3fc19bdedf1f48be8a, thank you!

PR #380, opened 2020-11-15
The minimum no of onion skin to the future and past is set to 0(Default)

abhinav3967:
If someone need to look at only the past they can do that and if someone has to look only into the future this would be needed


OverloadedOrama (approved):
Awesome, thanks!

PR #381, opened 2020-11-16
Setting the tilemode in each project different

abhinav3967:


OverloadedOrama:
Hmm, I think we might need more opinions on this one. For some people it might be more convenient to have the tile mode stay the same when switching tabs. I'll wait for more opinions before merging. In general, it might be best to start [issues](https://github.com/Orama-Interactive/Pixelorama/issues) before creating a pull request that is targeted to change how an already existing feature works, so people can offer their thoughts and opinions.

I took a quick look at the PR, it seems that tile mode drawing is always enabled, even if tile mode itself is set to "None". What I mean is, if you have tile mode to "None", you can still draw outside the canvas, just like you can when tile mode is enabled. Is this a bug?

OverloadedOrama (changes requested):
I tested it more. I found out a bug where, if you open a project or an image file and then you select a tile mode, the background gets tiled, but not the image itself. I suspect it has to do with the usage of arrays, although I haven't debugged it yet.

Nevertheless, I recommend avoiding arrays as much as possible. I did that mistake too in the past. Arrays are a great source of bugs and even crashes. In this scenario, what I would recommend is to have each project's tile mode be a member of the Project class. In other words, we could use a `tile_mode` variable in Project.gd, instead of relying on an array. And every time the user changes the tile mode, the `tile_mode` of Project.gd would change. This would make it easier to handle, potentially less problematic, and we could even save it to the .pxo file.

abhinav3967:
As there has been update in tile mode i am gonna make a new fork and PR, sorry for being late i had exam

Issue #271, opened 2020-06-25
Scrollbars for image panning


wyattbiker:
When canvas is larger than image viewport would be nice to allow scrollbars appear or provide some panning capability.

OverloadedOrama:
Panning is already possible with the middle mouse button, the Space button, and with the arrow keys on your keyboard. Scrollbars would be nice to have too.

wyattbiker:
The Space and Arrows keys work but need to pan faster. Also when panning, it tabs through various buttons in the Right panel.

As an FYI:
Mac doesn't have clickable middle button. It has Left click, Right click and touch slide ( acts like middle scroll wheel of mouse)  https://youtu.be/qn6fz_6_fcw?t=178

For example with ASE I can touch slide up/down/left/right and scrolls the viewport. Actually this happens for any application scroll bars appear including Pixelorama.

PR #390, opened 2020-12-04
Allow zooming with a touchpad

Schweini07:
Now through panning/zooming on a touchpad (e.g. from a latpop) it's possible to zoom.
I think I'll open another PR later so that panning moves the viewport (#271), let me know what you think.
The motions performed are "Panning with Inertia" and "Zoom" in this picture:
grafik


Schweini07:
I got moving the canvas through panning to work, should I replace the current panning zoom action with it?

Schweini07:
Looking through other image editing programs it seems like panning for moving the canvas and the zooming gesture for zooming is the standard. I'll open a new PR in a few minutes, with those changes.

PR #391, opened 2020-12-04
Allow moving the canvas and zooming with the touchpad

### LINKED ISSUE #271: Scrollbars for image panning

Schweini07:
Please look at #390 before, I explained the gestures there.
Now zooming trough panning is replaced with moving the canvas through panning.

Fixes #271(?)

### DIFF
--- a/src/UI/Canvas/CameraMovement.gd
+++ b/src/UI/Canvas/CameraMovement.gd
@@ -126,6 +126,13 @@ func _input(event : InputEvent) -> void:
 			zoom_camera(-1)
 		elif event.is_action_pressed("zoom_out"): # Wheel Down Event
 			zoom_camera(1)
+		elif event is InputEventMagnifyGesture: # Zoom Gesture on a Laptop touchpad
+			if event.factor < 1:
+				zoom_camera(1)
+			else:
+				zoom_camera(-1)
+		elif event is InputEventPanGesture: # Pan Gesture on a Latop touchpad
+			offset = offset + event.delta * zoom * 7 # for moving the canvas
 		elif event is InputEventMouseMotion && drag:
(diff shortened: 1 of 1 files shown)


OverloadedOrama:
I don't think it solves #271, since it's asking for scrollbars that can also be used with the mouse and this PR is only about touch controls. Still though, it looks nice and I'll test it later. Does it work for mobile devices (and touch displays in general) too?

Schweini07:
I think those Input commands are only for laptop touchpads,  but maybe it works?!

And the possible fix for #271 is because of this comment from the issue author:
_As an FYI:_
_Mac doesn't have clickble middle button. It has Left click, Right click and touch slide ( acts like middle scroll wheel of mouse)_ _https://youtu.be/qn6fz_6_fcw?t=178_

_For example with ASE I can touch slide up/down/left/right and scrolls the viewport. Actually this happens for any application_ _scroll bars appear including Pixelorama._

Schweini07:
Looking at the InputEventGesture class it states:
`Base class for touch control gestures.`
So it probably works on mobile.

OverloadedOrama (approved):
Okay! Looks good to me. Thank you!

OverloadedOrama:
Ah, I did some more research and looks like this change will only affect the macOS platform. https://github.com/godotengine/godot/issues/13139

There's an open PR in Godot to make them work for all platforms though, so let's cross our fingers that it'll get merged and wait for Godot 4.0 :D

Schweini07:
What a pity.
Well with luck waiting for Godot 3.2.4 or 3.2.5 it's implemented!

PR #393, opened 2020-12-05
Change shortcut key from CTRL to CMD on OSX

### LINKED ISSUE #67: Supporting MacOS hotkeys
It's not very comfortable to use Ctrl+* to shortcut in MacOS. It would be better if Pixelorama will support Cmd+*
_I'de like to fix it in this week_

Schweini07:
Note that the events in the InputMap still are like this: `ctrl + cmd + s`. Maybe you wanna change it to something like this: `ctrl + s` someday. Also thanks for the better solution!
Closes #67.


OverloadedOrama (approved):
Awesome, thank you!

PR #394, opened 2020-12-07
Improve Cpu Usage while in Idle

Schweini07:
Cpu Usage is now 1% while in Idle.
Needs some testing for usability tough, set for v0.8.3/0.9.

### DIFF
--- a/src/Main.gd
+++ b/src/Main.gd
@@ -121,8 +121,13 @@ func handle_backup() -> void:
 
 
 func _notification(what : int) -> void:
-	if what == MainLoop.NOTIFICATION_WM_QUIT_REQUEST: # Handle exit
-		show_quit_dialog()
+	match what:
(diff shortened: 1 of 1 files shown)


OverloadedOrama:
This is great! I tried it and the CPU usage when idle is indeed very noticeable. I'm a bit worried about the user experience though. The FPS is set to 1 when the mouse exits the window, which may confuse the users, especially when an animation is already playing. Perhaps it would be better if the behavior was changed to set the FPS to 1 when the window loses focus instead, and/or set this behavior toggleable in the Preferences.

Schweini07:
I though about that. I also think that would be better, I'll update soon.

Schweini07:
Done!

OverloadedOrama (approved):
Looks good, thank you!

KripC2160:
Really cool to see these features be added! I knew sometimes I would forget to close the program and my computer fans would be going crazy while I was doing other stuff in the background.

PR #398, opened 2020-12-09
Add move tool

Schweini07:


luiq54:
@Schweini07 Great work! Can you make it work on the second viewport as well (The one that you can drag in from the right edge)?

Issue #313, opened 2020-08-25
Move Tool


Just-Enix:
When I discovered Pixelorama I knew this was the perfect pixel art animating software I could find, and I like the UI too.

However, after I used all the tools and explored thro Pixelorama I found something quite strange, I cant cut what I select and I cant move what I draw.

I was motivated to start making the animations for my game but when I saw that there is not any way to move what you select I lost this motivation. This is something very useful since you can just move the sprites without redrawing everything one pixel to the left.

Please, I request a Move Tool. This will make my work much easier. 

Thanks.

Just-Enix:
Oh, I saw youre working on this, sorry for not paying attention.

OverloadedOrama:
Hello, you can move your selection by dragging it while holding Shift. In the future this will change though, most likely with a move tool as you suggest
Not entirely sure yet, I need to think a bit more about it.

Just-Enix:
> Hello, you can move your selection by dragging it while holding Shift. In the future this will change though, most likely with a move tool as you suggest
> Not entirely sure yet, I need to think a bit more about it.

Oh thx u so much for telling me how to drag pixels, maybe a move tool wont be needed, what you really need is tell the users how to do this bc there is no explicit info of this

PR #399, opened 2020-12-11
Add pan tool

Schweini07:
@luiq54 the second canvas is now dragable.

I also have to learn Git better. I needed one hour for finally putting this up and 1 minute working on it in Godot.


luiq54:
Looks great!

OverloadedOrama (changes requested):
So, functionality-wise it's very good! It will be a useful tool to have and it works well. However, I am mostly worried about the name of the tool, and its icon. A "move" tool usually refers to a tool that moves the pixels of the image (see #313) and not just the canvas. I think if we renamed it to "Pan" it would be a much better name. A hand icon, like this from Godot
image

also fits better for this purpose. I'd recommend using an icon like that for the Pan tool, and we can keep the one you made for a move tool that moves parts of the image.

If you could make these two changes, and if you agree of course, then I think it's ready to get merged. If you have issues making a hand icon, we could ask @Erevoid to handle that.


(Also, it's best if you avoided including files to the commit that don't have any relevant changes, like Main.tscn)

Schweini07:
Thanks for the feedback!
Regarding Main.tscn, I had a lot of problems with Git, but maybe I am able to get it right this time.

Schweini07:
I did the renaming, but I don't think I will get a good hand icon done.
I would have nothing against @Erevoid  making those.
Should I already push, with the old icons, or wait?

OverloadedOrama:
> I did the renaming, but I don't think I will get a good hand icon done.
> I would have nothing against @Erevoid making those.
> Should I already push, with the old icons, or wait?

Go ahead. You can push the changes and we can merge the PR now, and change the icon later.

Schweini07:
Done!

OverloadedOrama (approved):
Great, thank you!

PR #405, opened 2020-12-19
Use undo and redo while holding the respected shortcuts

### LINKED ISSUE #401: Key combinations don't repeat when held
**Pixelorama version:**
v0.8.2-stable

Arch Linux 5.9.11.arch2 i3/Nvidia 1060 3GB drivers version 455.45.01

**Issue description:**
While holding ctrl+z, ctrl+y or any other key combination, their respective action gets executed only once.
It get's tiring really fast to keep pressing the same combination over and over again.

**Steps to reproduce:**
Draw a bunch of lines, then hold ctrl+z. Notice how only the last action gets reverted, and nothing more.

If this isn't intended behaviour, I can send a PR myself.

Schweini07:
Closes #401.

As you may noticed, the code below `if !event.is_echo()` is only executed when `is_echo` is true. I think this is fine, but maybe it should be done differently?

### DIFF
--- a/src/Main.gd
+++ b/src/Main.gd
@@ -55,10 +55,26 @@ func _input(event : InputEvent) -> void:
 		if get_focus_owner() is LineEdit:
 			get_focus_owner().release_focus()
 
-	if event.is_action_pressed("redo_secondary"): # Shift + Ctrl + Z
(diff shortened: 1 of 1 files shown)


OverloadedOrama (approved):
Looks great, thank you!

PR #417, opened 2021-01-03
Fix tiling mode opacity by predrawing current frame

### LINKED ISSUE #416: Tile mode opacity and per layer opacity are applied at the same time
**Pixelorama version:** [5f56cb523f3766a9ddf906003f150f95c65c098e](https://github.com/Orama-Interactive/Pixelorama/tree/5f56cb523f3766a9ddf906003f150f95c65c098e)

**Issue description:** In #414 I've introduced new bug by applying layer opacity and tile mode opacity at the same time. Instead per layer opacity should be applied first, then tile mode opacity on the result.

**Example:**
For 50% tile mode opacity and 100% both layers opacity red part from the layer below shouldn't be visible:
Godot_v3 2 3-stable_win64_ZnWtQE9hBI

kleonc:
Fixes #416.

- Now the current frame is drawn to a separate Viewport (with transparent background) taking into account only per layer opacity,
- tiling is drawn by drawing current frame's ViewportTexture with tile mode opacity applied (using premultiply alpha blending).

### DIFF
--- a/src/UI/Canvas/Canvas.gd
+++ b/src/UI/Canvas/Canvas.gd
@@ -9,6 +9,8 @@ var can_undo := true
 var cursor_image_has_changed := false
 var sprite_changed_this_frame := false # for optimization purposes
 
+onready var currently_visible_frame : Viewport = $CurrentlyVisibleFrame
+onready var current_frame_drawer = $CurrentlyVisibleFrame/CurrentFrameDrawer
 onready var grid = $Grid
 onready var tile_mode = $TileMode
 onready var indicators = $Indicators
@@ -42,6 +44,8 @@ func _draw() -> void:
 
 	if Global.onion_skinning:
 		onion_skinning()
+	currently_visible_frame.size = Global.current_project.size
+	current_frame_drawer.update()
 	tile_mode.update()
 	draw_set_transform(position, rotation, scale)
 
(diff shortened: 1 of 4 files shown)


kleonc:
@OverloadedOrama, I'm not sure where requests for updating (redrawing) current frame should be applied (same about updating its `Viewport` size). As you can see in the code currently I'm doing it every frame (in `_process`) which for sure is not needed.
Also since current frame is now rendered as a single texture it might be potentially used in previews.

OverloadedOrama:
I would suggest to remove the _process() method completely, and instead do the update (redraw & size changing) on Canvas.gd's update_texture() method and Project.gd's change_project() method.

This seems to be working, inside the update_texture() method.
```
$CurrentlyVisibleFrame.size = project.size
$CurrentlyVisibleFrame/CurrentFrameDrawer.update()
```
I suggest using variables for these node paths for readability. These 2 lines should also go to Project's change_project() method, and it should be working fine.

kleonc:
> instead do the update (redraw & size changing) on Canvas.gd's update_texture() method and Project.gd's change_project() method.

It wouldn't be enough. For example it wouldn't be updated when another frame is selected or when layer's opacity is changed. However updating it only in `Canvas._draw()` seems to work fine.

OverloadedOrama:
Yeah you are right, Canvas._draw() seems to be a better option, and from what I tested, we also don't need to put it in Project.change_project() either.

kleonc:
> I suggest using variables for these node paths for readability.

I forgot about that part, gonna fix it in a moment.

kleonc:
Seems fine to merge.

OverloadedOrama (approved):
Thank you!

PR #420, opened 2021-01-04
Update GUI for Cel.opacity on frame changed instead of on layer change

### LINKED ISSUE #419: Current cel's opacity isn't always properly updated in the GUI
**Pixelorama version:** [92a22fe9bc05b3d1e70ae2e49b47f80da90c9672
](https://github.com/Orama-Interactive/Pixelorama/tree/92a22fe9bc05b3d1e70ae2e49b47f80da90c9672)

**Issue description:** Title.

**Steps to reproduce** (just one of many ways):

1. Create new project.
2. Change opacity of the current cel. Example result:
Godot_v3 2 3-stable_win64_KqZsSFGlTA

3. Add a new frame. Opacity of the previously selected cel will be shown (incorrect). Example result: 8jRvVSnT73

4. Click newly added frame. Opacity of the currently selected cel will be shown (correct). Example result: T1PDohwMDq

kleonc:
It's `Cel`'s property and thus it changes when `frame_changed` is called (also `layer_changed` calls `frame_changed` on its own). Fixes #419.

### DIFF
--- a/src/Classes/Project.gd
+++ b/src/Classes/Project.gd
@@ -465,15 +465,16 @@ func frame_changed(value : int) -> void:
 	Global.disable_button(Global.move_left_frame_button, frames.size() == 1 or current_frame == 0)
 	Global.disable_button(Global.move_right_frame_button, frames.size() == 1 or current_frame == frames.size() - 1)
 
+	if current_frame < frames.size():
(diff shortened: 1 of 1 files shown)


OverloadedOrama (approved):
Nice find! Really appreciate the detailed bug reports, as well as the fixes. Thank you!

PR #425, opened 2021-01-06
Remove unneeded Canvas.location, Grid.location, TileMode.location

kleonc:
All of these are initialized to `Vector.ZERO` and they're never changed. Also all of these classes (`Canvas`, `Grid`, `TileMode`) extend `Node2D` and thus each of them has a `transform` and a `position` (which is for example used instead of `location` when drawing mirrored view).


OverloadedOrama (approved):
Thanks!

PR #378, opened 2020-11-12
Improved the tiling feature

abhinav3967:
I have added more tiling options like tile in x axis, y axis and both axis together.

PR #427, opened 2021-01-09
Draw Grid only over Canvas. Added PixelGrid.

### LINKED ISSUE #426: Grid is not properly drawn in tile mode when project size is not divisible by grid size
**Pixelorama version:** [b1c0344d0ba5dcb974fab862a25467888adf1cef](https://github.com/Orama-Interactive/Pixelorama/tree/b1c0344d0ba5dcb974fab862a25467888adf1cef)

**Issue description:** Title. Grid origin should stay at the Canvas origin.

**Example:** 
- Grid size divisible by project size:
NRakMywkIf

TileMode.NONE (ok):
s1fXbRzZVJ

TileMode.BOTH (ok):
Godot_v3 2 3-stable_win64_rXQ6y36LHr

- Grid size not divisible by project size:
iNqPXZ6kCU
TileMode.NONE (ok):
Godot_v3 2 3-stable_win64_uIhxuIeHID
TileMode.BOTH (incorrect):
aaIGtSGYop

**Suggested solution:** Literally copying how Krita does it: drawing grid only over original Canvas as it doesn't make much sense to draw it over the tiling area because of issues like this one (it's not guaranteed to tile seamlessly). Additionally add option to draw per pixel grid when zoomed besides some threshold (which will be drawn everywhere when in tile mode).

kleonc:
Fixes #426 (implemenetation of the solution I suggested in there).
- Draw Grid only over the Canvas (when in tiling mode).
- Added PixelGrid.
- Pixel Grid is shown only when it's enabled and camera is zoomed close enough.
- Settings (to distinguish between Grid and PixelGrid default grid_width, grid_height settings are changed):

tMFQfBS6Dw
- Default values might need changing.
- Now both Grid and PixelGrid are drawn above (after) TileMode. Grid is drawn above (after) PixelGrid.

<details>
<summary>Show images</summary>

**Previous look:** see #426.
**Current look:** (both Grid and PixelGrid are enabled everywhere)

NRakMywkIf
Godot_v3 2 3-stable_win64_EFy7s9MPPr
Godot_v3 2 3-stable_win64_Uk76mUXdzj
Godot_v3 2 3-stable_win64_lHi4f4RVzT
vAuBEgqiZL

iNqPXZ6kCU
48n2TyZCCQ

Godot_v3 2 3-stable_win64_zLFjdbta6u
u2NWbpp6uu
Godot_v3 2 3-stable_win64_nKZ9gT25y3

</details>

### DIFF
--- a/Translations/Translations.pot
+++ b/Translations/Translations.pot
@@ -148,6 +148,9 @@ msgstr ""
 msgid "Show Grid"
 msgstr ""
 
+msgid "Show Pixel Grid"
+msgstr ""
+
 msgid "Show Rulers"
 msgstr ""
 
@@ -814,6 +817,21 @@ msgid "Grid color:"
 msgstr ""
 
 msgid "A color of the grid"
+msgstr ""
+
+msgid "Pixel Grid"
+msgstr ""
+
+msgid "Show at zoom:"
+msgstr ""
+
+msgid "Sets the minimal zoom at which pixel grid will be shown"
+msgstr ""
+
+msgid "Pixel grid color:"
+msgstr ""
+
+msgid "A color of the pixel grid"
 msgstr ""
 
 msgid "Transparency"
--- a/project.godot
+++ b/project.godot
@@ -429,6 +429,11 @@ right_pan_tool={
 "events": [ Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":0,"alt":true,"shift":false,"control":false,"meta":false,"command":false,"pressed":false,"scancode":77,"unicode":0,"echo":false,"script":null)
  ]
 }
+show_pixel_grid={
+"deadzone": 0.5,
+"events": [ Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":0,"alt":false,"shift":false,"control":true,"meta":false,"command":true,"pressed":false,"scancode":72,"unicode":0,"echo":false,"script":null)
+ ]
+}
 
 [locale]
 
--- a/src/Autoload/Global.gd
+++ b/src/Autoload/Global.gd
@@ -52,10 +52,12 @@ var default_image_width := 64
 var default_image_height := 64
 var default_fill_color := Color(0, 0, 0, 0)
 var grid_type = Grid_Types.CARTESIAN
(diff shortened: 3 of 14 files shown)


kleonc:
@OverloadedOrama, I didn't do anything about translations though. Not sure which strings and to which file should I add. You can check added strings in general (like if tooltips are ok etc.)
Also I'd like to hear suggestions for default values, current ones (colors) aren't too distinguishable when both grids are showed.

OverloadedOrama:
> @OverloadedOrama, I didn't do anything about translations though. Not sure which strings and to which file should I add. You can check added strings in general (like if tooltips are ok etc.)
> Also I'd like to hear suggestions for default values, current ones (colors) aren't too distinguishable when both grids are showed.
> And I didn't touch isometric grid yet (but it's also being drawn only over the Canvas). But that's for another fix / PR.

I think the tooltips and default values are okay. As for translations, all you have to do it put the newly added strings in the Translations.pot file. Crowdin handles the rest.

OverloadedOrama (approved):
It looks good! Really appreciate the enums you added in TopMenuContainer too.

I think the grid being drawn in tile mode was added in #378, so I'd like @abhinav3967's opinion about this too.

kleonc:
> As for translations, all you have to do it put the newly added strings in the Translations.pot file. Crowdin handles the rest.

Done.


> I think the grid being drawn in tile mode was added in #378, so I'd like @abhinav3967's opinion about this too.

Sure thing. I'm not sure about this one too, maybe there's a better solution. I've just noticed that grid sometimes can't both be shown the same over each tile and tile seamlessly on its own.
<details>
<summary>When showing the same over each tile you might end up with something like this</summary>

Godot_v3 2 3-stable_win64_UFDydb6QB2
</details>
<details>
<summary>When tiling grid seamlessly on its own you might end up with something like this</summary>

aaIGtSGYop

</details>
Not sure if any of these behaviours are wanted.


Besides that adding a setting for a grid offset is probably a good idea.

OverloadedOrama (approved):
I think this is okay to get merged, and we can see if we can find an alternative solution to #426 in the future, but this will do for now. Thank you!

PR #430, opened 2021-01-16
Refactored isometric grid

kleonc:
Now it should be drawn properly in the target rect.
Settings (in pixels):
- cell's AABB size,
- offset from the canvas origin.

Few images with additional debug draw:
sbTJwd7E6r
V9ANrbEeqI
5qrwpoJ3LG
2Vzfi0MBbZ

PR #434, opened 2021-01-17
One shared offset for rectangular and isometric grids, added "Draw Grid over Tile Mode" toggle

kleonc:
Based on #430, should be merged after. I think I'm done with grids for now.

Makes drawing grid only over the canvas (introduced in #427) toggleable. This makes grid drawing like in #426 possible again (with custom grid offset).


OverloadedOrama:
I merged #430 first but it seems to have created some conflicting files with this PR now.

kleonc:
Rebased, should be fine now.

OverloadedOrama:
Looks good and works well. I am wondering however if we need different offset options for the rectangular and isometric grids. Is there an advantage to have them as different options? From what I see the only difference it makes is when both grids are enabled, but I don't know if there's a practical reason not to have them have the same offset, but I could be wrong of course.

I'm okay with them being different options, but I'm afraid it might result in users having to do extra steps that can be avoided. Opinions are of course welcome.

kleonc:
I was wondering about the same thing. Different offsets wouldn't make sense if there was no option to draw both rectangular and isometric grids at the same time. But such option is available and thus I've made offsets separated because it gives a little more flexibility if someone needs it, that's all. I didn't have any particular use case in mind.

However, as you said, it might add some extra steps for a regular user. Also I don't think that drawing both grids will be used often (if ever). Hence I'm perfectly fine to make a single offset for both grids instead.

It's also probably worth mentioning that origin of isometric grid in its current implementation defines where the middle point of a cell is, not where the grid lines intersect. For a rectangular grid origin defines where the lines intersect. I've implemented isometric grid origin like this beacuse it seemed more practical to me (also it kinda preserved the default look of isometric grid before my changes). I can change it if that's not preferred behaviour:

<details>
<summary>Current implementation: isometric grid origin defines center of the cell</summary>

Godot_v3 2 3-stable_win64_FyG6aTIspZ
Godot_v3 2 3-stable_win64_bW82lDqqHp

</details>
<details>
<summary>Different approach: isometric grid origin defines lines intersection point</summary>

Godot_v3 2 3-stable_win64_DBQABqCQXp
Godot_v3 2 3-stable_win64_kAnOkQNTZ0


</details>

OverloadedOrama:
I like the current implementation of the origin of the isometric grid, so I'd say to leave that as it is. As for the rectangular and isometric offsets, I think it would be better if they were the same. Since you're fine with that, you could make this change and the PR will be ready for merge.

kleonc:
That's how settings look now:
Godot_v3 2 3-stable_win64_1TYvw03MlB

Should be fine to merge.

OverloadedOrama (approved):
Thank you!

PR #439, opened 2021-01-20
Speedup Draw.update_mask()

kleonc:
`Draw.update_mask()` is slow and it's being called by tools (pencil etc.) every time on `draw_start()`. This PR makes it a little faster by using the fact that [Array.resize()](https://docs.godotengine.org/en/3.2/classes/class_array.html?highlight=array#class-array-method-resize) makes new elements null contrary to [PoolByteArray.resize()](https://docs.godotengine.org/en/3.2/classes/class_poolbytearray.html#class-poolbytearray-method-resize) which isn't zeroing new elements.
<details>
<summary>PoolByteArray created by passing in Array with nulls is indeed zeroed.</summary>

```
var n := 10
var a := PoolByteArray()
a.resize(n)
print(a.hex_encode())
for i in range(n):
    a[i] = 0
print(a.hex_encode())

var arr := []
arr.resize(n)
print(arr)
var b := PoolByteArray(arr)
print(b.hex_encode())
```
example output:
```
7200650073003a002f00
00000000000000000000
[Null, Null, Null, Null, Null, Null, Null, Null, Null, Null]
00000000000000000000
```
</details>

---
Example comparison of execution times for 1920x1080 canvas size (calls are trigerred by single clicks with pencil tool):
kVFCHsJbi0

It's still not good but it's better.

### DIFF
--- a/src/Tools/Draw.gd
+++ b/src/Tools/Draw.gd
@@ -131,10 +131,10 @@ func update_mirror_brush() -> void:
 
 func update_mask() -> void:
 	var size := _get_draw_image().get_size()
-	_mask = PoolByteArray()
-	_mask.resize(size.x * size.y)
-	for i in _mask.size():
-		_mask[i] = 0
+	# Faster then zeroing PoolByteArray directly. See: https://github.com/Orama-Interactive/Pixelorama/pull/439
+	var nulled_array := []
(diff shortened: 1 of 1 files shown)


OverloadedOrama (approved):
Amazing, thank you!

PR #443, opened 2021-01-28
Cache tile mode rects

kleonc:
When drawing in a tile mode `get_tile_mode_rect` is being called as many times as `_set_pixel`. This PR make tile mode rects being cached for a little speedup.

---
Example performance comparison for 128x128 canvas size and single clicks with 100x100 pencil tool:

- Before:
9H3pH8lyEc

- After:
eAhZrX015f


OverloadedOrama (approved):
Thank you!

PR #444, opened 2021-01-29
Var transparent mode

Variable-ind:
I modified my earlier version a little bit so that it would reset to zero in full screen mode hence removing the blackness issue


kleonc:
For sure you're not blending colors correctly:
<details>
<summary>Opaque image (100% transparent mode)</summary>

yZdInjYGI6
</details>

<details>
<summary>Not opaque image (100% transparent mode) </summary>

#### It's too dark, blending with the white background shouldn't make it darker.
NZNlnAZWzn
</details>

---
Setting blending mode to premultiplied alpha for the main ViewportContainer seems to fix it:
<details>
<summary>Not opaque image (100% transparent mode).</summary>

EqRnzyV5uG
</details>

Variable-ind:
it is mot a bug. If you look at the code, the transparency does not affect the image(even a little bit), it only effects the checker shader and the background.

Variable-ind:
however it does seem to get the job done so i am going to add it

kleonc:
> it is not a bug. If you look at the code, the transparency does not affect the image(even a little bit), it only effects the TransparentChecker.shader and the background.

Image can be transparent and thus its colors are already multiplied by alpha of the layers etc. So if you're changing Viewport's background to be transparent you need to take into account that the alpha was already applied to the image (RGB values of the image are already multiplied by the alpha). Otherwise you'll end up with the darker colors when `backgroundAlpha != 1.0` because RGB values of the image would be again multiplied by some factor < 1.0 (which you can see happening on the image in my previous comment).

Variable-ind:
yes i found it just now and am enabling material.blend_mode = CanvasItemMaterial.BLEND_MODE_PREMULT_ALPHA thanks for letting me know

src/Main.gd
```diff
@@ -7,6 +7,16 @@ var is_quitting_on_save := false
 
 # Called when the node enters the scene tree for the first time.
 func _ready() -> void:
+	var Alternate_transparent_Background = ColorRect.new()
```
OverloadedOrama, line 10: Rather minor, but make sure variable names are in snake_case, according to [GDScript style guide](https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_styleguide.html#functions-and-variables). So the variable name should be alternate_transparent_background

src/Main.gd
```diff
@@ -7,6 +7,16 @@ var is_quitting_on_save := false
...
+	add_child(Alternate_transparent_Background)
+	move_child(Alternate_transparent_Background,0)
+	Alternate_transparent_Background.visible = false
+	Alternate_transparent_Background.name = "Alternate transparent Background"
```
OverloadedOrama, line 14: Same as the above comment, node names should be in PascalCase, according to the [GDScript style guide](https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_styleguide.html#classes-and-nodes). The node's name should be AlternateTransparentBackground.

OverloadedOrama (changes requested):
It seems to be working well, good job! I left some comments too, and I would also like to propose to change "Transparency Mode" to "Window Transparency", in order to make what the option does more obvious to the user. As for the option's UI, I'm still unsure if this is a good place for it, but I suppose we could leave it as it is for now, and we could change it in the future if we think of something better.

src/Main.gd
Variable-ind: Done

src/Main.gd
Variable-ind: Done

OverloadedOrama (approved):
All right, awesome, thank you!

Issue #445, opened 2021-01-29
Crash when reloading more than one backuped empty project


kleonc:
**Pixelorama version:** [4001a32f2b2badc3cd31cd49e09ebbf4d8378473](https://github.com/Orama-Interactive/Pixelorama/tree/4001a32f2b2badc3cd31cd49e09ebbf4d8378473)

**Issue description:** Title.
Seems like `OpenSave`'s member arrays: `current_save_paths` and `backup_save_paths` are assumed to be the same size as arguments passed to the `OpenSave.reload_backup_file` method which seems to not always be the case.
![Godot_v3 2 3-stable_win64_7pmZsbo1A5](https://user-images.githubusercontent.com/9283098/106285066-4d5a6500-6244-11eb-9eb5-b5f9ed6ac7a2.png)

**Steps to reproduce:**

1. Create at least 2 empty projects (with nothing being drawn) and wait for each of them be autosaved.
2. Terminate the program (for example by clicking stop button in the Godot). Closing Pixelorama with the X button seems to not cause the crash.
3. Run Pixelorama again and confirm reloading backups.

OverloadedOrama:
Seems like the issue happens because importing a new project when the current one is empty, replaces the empty project instead of creating a new tab. This behavior happens when importing backups too, the first project is empty, so the second one replaces it, thus a new value is not being appended in `backup_save_paths[]`, which results in the error. It should get fixed in the next commit.

PR #446, opened 2021-01-29
Fix potential index out of bounds error

kleonc:
In `OpenSave.reload_backup_file` after the first removal with `project_paths.remove(backup_paths.find(backup))` the index returned by `backup_paths.find(backup)` will no longer point to the project path respective to that `backup`, `project_paths` and `backup_paths` arrays would be "desynchronized". This PR makes both arrays be changed properly when iterating over the backup paths (and shrinked if needed after the loop).

### DIFF
--- a/src/Autoload/OpenSave.gd
+++ b/src/Autoload/OpenSave.gd
@@ -529,19 +529,21 @@ func remove_backup_by_path(project_path : String, backup_path : String) -> void:
 
 
 func reload_backup_file(project_paths : Array, backup_paths : Array) -> void:
+	assert(project_paths.size() == backup_paths.size())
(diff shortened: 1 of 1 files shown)


OverloadedOrama (approved):
Looks good! This closes #445 right?

kleonc:
> 
> This closes #445 right?

No, it's unrelated. Just noticed this one after a brief look at the `OpenSave.reload_backup_file` method when I've got crash described in #445. But error fixed here isn't the cause of #445.

OverloadedOrama:
Oh you are right. I tested the GitHub Actions generated build first which doesn't crash and I thought the issue was fixed, my bad. Anyway, thank you!

Issue #356, opened 2020-10-16
Enhance Palette panel/editor


novhack:
Currently editing of a palette is done in a dialog. Most of other software allows users to add and change colors directly from the main UI which is in my opinion a better workflow. I propose these:

Changes:

- [x] Plus/Minus buttons to add/delete colors in a Palettes panel. Left click on Plus button adds left tool color, right click adds right click color to the end of the palette.
- [x] Double click on a color opens color editor dialog.
- [x] Remove delete palette button from panel and move it to edit palette dialog. It's a feature that is most likely not used a lot.
- [x] Move and copy colors directly in the panel.
- [x] Drag and drop on existing color = swap colors.
- [x] Drag + `shift` = insert color in between two existing colors.

Additions:
- [x] Ability to Change palette grid size.
- [x] Organize colors freely on the palette grid (with empty spaces).
- [x] Drag + `ctrl` = copy color to empty grid space.

Future improvements:
- [x] Replace current JSON palettes implementation with TRES.
- [ ] Add palette modifications to undo/redo system.
- [ ] Import palettes based on file headers not file extension. Files like `palette.gpl.txt` should be imported correctly.
- [ ] Indexed colors mode. Changing a color in the palette changes the same color in the canvas. #171
- [ ] Colors multi select
- Ability to select multiple colors and use scroll wheel to iterate through them.
- [ ] Colors sorting.
- [ ] Color gradients generation.
- [ ] Select multiple colors and generate one color that has their average value. This workflow can help with cleaning up images. [Hue Shifting in Pixel Art (Color Tutorial)](https://www.youtube.com/watch?v=PNtMAxYaGyg)
- [ ] Support for project palettes. Embedded into .pxo file for better portability.
- [ ] Merge palettes
- A dialog with check boxes that allows user to merge all colors from other palettes to a current palette

novhack:
Issue #187 was already suggesting some of the changes I described.

PR #447, opened 2021-01-30
Replace old palette system with a new one

novhack:
I completely remade the palette system. Basically all features from #356 are implemented except undo/redo support.

This pull request breaks compatibility with old palettes. When I was redesigning the load/save system I decided to go with Resource format for palettes so the old JSON won't load anymore. Keeping legacy JSON related code seemed like a bad idea.

image
New dialogs:
image
image


kleonc:
Issues (not sure if related to undo/redo being not implemented yet):

<details>
<summary> GUI is not being instantly updated when modyfying palettes loaded from the disk (not created at the current Pixelorama run). </summary>

#### Examples:
1. Drag and drop:
Godot_v3 2 3-stable_win64_SrjKZRXRIA
2.  Add / remove colors:
Bk05LKWNOK

</details>


<details>
<summary> Editing palette loaded from the disk (not created at the current Pixelorama run) duplicates the entry in the palettes dropdown. </summary>

kVpm5SAOHU

</details>

novhack:
Thank you for testing. What operating system do you use? I see that the path for the palette is in res:// which doesn't happen to me - all imported palettes should be stored/copied to appdata.

kleonc:
Windows 7. I was running [this version](https://github.com/novhack/Pixelorama/tree/6fc801c3eb19bdf7ba836d0fb9398f6588b4f3ce) in debug mode, right from the Godot.
<details><summary> All I have in appdata are: a backup, config file and some logs. </summary> 

```
C:\Users\MyUsername\AppData\Roaming\Godot\app_userdata\Pixelorama\backup-1609273475-0
C:\Users\MyUsername\AppData\Roaming\Godot\app_userdata\Pixelorama\cache.ini
C:\Users\MyUsername\AppData\Roaming\Godot\app_userdata\Pixelorama\logs
C:\Users\MyUsername\AppData\Roaming\Godot\app_userdata\Pixelorama\logs\godot.log
C:\Users\MyUsername\AppData\Roaming\Godot\app_userdata\Pixelorama\logs\godot_2021-01-31_13-04-56.log
C:\Users\MyUsername\AppData\Roaming\Godot\app_userdata\Pixelorama\logs\godot_2021-01-31_13-05-38.log
C:\Users\MyUsername\AppData\Roaming\Godot\app_userdata\Pixelorama\logs\godot_2021-01-31_13-09-35.log
C:\Users\MyUsername\AppData\Roaming\Godot\app_userdata\Pixelorama\logs\godot_2021-01-31_14-20-33.log
```
</details>

I've just downloaded current official version (`Pixelorama [Windows 64-bit].zip`) from [the site](https://orama-interactive.itch.io/pixelorama) and it has `pixelorama_data` directory right next to the `.exe` and that's where palettes are being added/edited. Same happens in your version: it's indeed modyfying contents of `./pixelorama_data/Palettes/`.

Erevoid (commented):
Eyo!

I tested your PR, I really like the new palette system in general. I'd like to comment some details I figured out while playing with the settings:
1. When you put high values on width and height it can crash the program. For example I put 100 for both width and height, and Pixelorama froze for several seconds. When I tried to add a color, it crashed. Μaybe have a limitation as to how many colours can be added with relatively no problem in most devices?
2. You can have 0 as the value for width or height, if there is no specific reason for that, why not have a minimum value of 1?
3. When you select a colour from the palette and then you switch the colour from the colour picker (not the palette's colour), the palette colour remains selected. That's a bit confusing since you have a specific colour shown at the colour picker and a different one selected at the palette, so I suggest deselecting the palette colour when the user changes the colour of the colour picker.
4. When the user creates a new palette it's not clear enough that a name is required so that the Ok button can be pressed. That can be fixed by adding some red text indicating that the name is missing when the field is empty

novhack:
At this point all the issues kleonc and Erevoid had should be resolved.

Palette size now has no impact on performance.
It's possible to scroll with a middle button in the palette panel.
There is now also an importer for "old" json palette format for backwards compatibility.

novhack:
I added two small fixes based on luiq54's feedback on Discord.

Erevoid:
Didn't find any new issues after the changes, it's looking good in my opinion!

OverloadedOrama (approved):
I think it looks good too. I've been testing it and it seems to be working well, so I'll go ahead and merge it now. Thank you!

PR #450, opened 2021-02-03
Added more standart devices from 8/16Bit era

THWLF:
I added a few new templates from other 8/16bit devices into the basic templates.


OverloadedOrama (approved):
Thank you!

PR #453, opened 2021-02-05
"SpriteSheet as layer" and "Replace Frame"  import options

Variable-ind:
This will allow us to use "SpriteSheet as layer" and "Replace Frame"  import options in pixelorama

Maintainer edit: This implements #448.


src/UI/Dialogs/PreviewDialog.gd
```diff
@@ -14,19 +14,44 @@ var brush_type : int = BrushTypes.FILE
...
+		if i == ImageImportOptions.REPLACE_FRAME:
+			import_options.add_item("Replace Frame")
+		if i == ImageImportOptions.NEW_LAYER:
+			import_options.add_item("New layer")
+		if i == ImageImportOptions.PALETTE:
+			import_options.add_item("New palette")
+		if i == ImageImportOptions.BRUSH:
+			import_options.add_item("New brush")
+		if i == ImageImportOptions.PATTERN:
+			import_options.add_item("New pattern")
```
OverloadedOrama, line 48: Out of curiosity, why did you add the items here with code instead of in the node inside PreviewScene.tscn? Anyway, if we are to keep them here, the for loop and the if statements are not needed, simply using `add_item()` will add all the items in order.

OverloadedOrama (commented):
It seems to be working well. The only issue I found is that if you switch between Spritesheet (new tab) and Spritesheet (new layer), is that the preview lines are not being updated. If you switch between new tab and new layer and you have different numbers of frames, the blue lines that separate the image are not getting updated.

I also left some comments in the code :)

src/Autoload/OpenSave.gd
```diff
@@ -388,6 +388,69 @@ func open_image_as_spritesheet(path : String, image : Image, horizontal : int, v
...
+		for xx in range(horizontal):
+			var cropped_image := Image.new()
+			cropped_image = image.get_rect(Rect2(frame_width * xx, frame_height * yy, frame_width, frame_height))
+			image_no += 1
+			if (start_frame + (image_no - 1)) < Global.current_project.frames.size():
+				# if frames are already present then fill those first
+				if image_no == 1:
+					open_image_as_new_layer(cropped_image, file_name, start_frame + image_no - 1)
+				else:
+					open_image_at_frame(cropped_image, Global.current_project.layers.size() - 1, start_frame + (image_no - 1))
```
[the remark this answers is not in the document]
Variable-ind: i placed it here so we would have if image_no == 1 (Since it's the 1st image and not the 0th image)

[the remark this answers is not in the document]
OverloadedOrama: I think having image_no == 0 would make more sense since in programming we're used to refer to the first item (in arrays, usually) with the number zero. We would also avoid doing unnecessary subtraction operations.

src/UI/Dialogs/PreviewDialog.gd
Variable-ind: i added them here because there is a known issue about the "OptionButton" in godot that if we can only add new options at the end. if we want to add an option at a selected position, we would have to remove all items and add them again

OverloadedOrama: Oh okay, good thinking. We should keep them there then, but we should remove the for loop and the if statements.

Variable-ind: ok

Variable-ind: yup, fixed it

src/Autoload/OpenSave.gd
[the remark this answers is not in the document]
Variable-ind: Done

OverloadedOrama (approved):
Great, thank you!

PR #455, opened 2021-02-09
Split image

Variable-ind:
Added a split-image function. it gets the overall image of a frame and splits that image


OverloadedOrama (changes requested):
I'm a bit confused; why does this PR add more fonts? They don't seem to be used anywhere.

Apart from that, the split image functionality seems to be working well, but undoing the changes seem to be taking more steps than necessary. Undoing one frame takes two steps (undo: change frame order and undo: add frame). I think it would be better if a single undo removed all of the frames added from split image, and a single redo would re-add them all.

Variable-ind:
i removed the additional fonts.
As i borrowed the code from "OpenSave.open_image_as_new_frame()" and "open_image_at_frame()" to create split image functionality so the undo comes from them as a side effect.

Issue #359, opened 2020-10-19
Circle and Rectangle Brush


cagdasc:
**Feature/enhancement description:**
Are there any circle and rect brush? I did lots of search but can not found.

novhack:
I think that both are planned as dedicated tools. Currently you can select a circle brush in pencil settings. It's coded in a way that if you scale it up it will still retain 1px outline.

cagdasc:
@novhack thank you but if we can draw free-form like drawing line press holding shift, it will be useful.

Issue #309, opened 2020-08-15
Shortcut Improvements


OverloadedOrama:
**Feature/enhancement description:**

- [x] Solve the macOS hotkey issue (#67).
- [x] Holding Ctrl+Z to undo many changes quickly. (#405)
- [ ] Add missing shortcuts for manipulating layers, frames, palettes and whatever else is missing. I might edit this to add a proper list of the missing shortcuts, feel free to give ideas!
- [ ] Add the ability to change shortcut bindings. Currently only implemented for tools and color switch, but it would be nice to be able to modify all shortcuts. Perhaps we should make it that the user can modify the entire (or most) of the Input Map bindings found in the project settings. This should also include the shortcuts used for each tool, for example Pencil and Shift to draw lines.
- [ ] Clear shortcut button. Not all shortcuts have to have a binding, the ability to clear a shortcut binding would be handy.
- [ ] Let the users create their own presets. I'm not sure how useful this is, but since we already use a Preset system, we might as well let the user save their changes as a preset and easily switch between them, like different "profiles".

The list may be edited to include more ideas in the future.

Blixtmen:
Being able to hold ctrl+z to undo many changes quickly is something that is missed from other softwares!

PR #456, opened 2021-02-15
Add Shapes Tool (WIP)

KawanWeege:
Addresses #359 

This PR adds a tool that allows you to draw shapes, like rectangles and elipses. It also allows you to draw shapes with a 1:1 aspect ratio (so you can draw circles and squares), by holding down the shift key.

Currently WIP, it still needs:
- Cleaning up
- Add the ability to move the shape's center to the click origin by pressing CTRL

I'll add images and gifs of the tool in action later today.


Variable-ind:
Ok so i checked and it works great but i noticed two things:
firstly the preview rectangle does not appear if the tool is selected with right mouse button.
and secondly, the preview rectangle does not update as the size of the shape brush is changed you might want to change that

KawanWeege:
> Ok so i checked and it works great but i noticed two things:
> firstly the preview rectangle does not appear if the tool is selected with right mouse button.
> and secondly, the preview rectangle does not update as the size of the shape brush is changed you might want to change that

I have addressed these two points in the latest commits.
During my tests with the shape thickness slider, I've noticed that sometimes drawing the shape will spam the console with errors, and I haven't found a consistent way of replicating it.
But other than that, the tool is mostly done, unless there is something I need to implement or I have forgotten to do something

Variable-ind:
OK, so i checked and it appears to work fine but an idea hit me,
what if we draw the preview rectangle using "Shift + click" and then paint using a simple click. i mean the idea is to be able to make the preview rectangle follow the mouse (similar to the preview of the pencil tool) and paint the preview shape whenever we click

KawanWeege:
> OK, so i checked and it appears to work fine but an idea hit me,
> what if we draw the preview rectangle using "Shift + click" and then paint using a simple click. i mean the idea is to be able to make the preview rectangle follow the mouse (similar to the preview of the pencil tool) and paint the preview shape whenever we click (if the rectangle belongs to the right-tool then right-click and vice versa)

I don't know if I understood correctly what you were asking for, but give a check on the commit I've just pushed. Press CTRL before releasing the mouse button to keep the preview on screen. Then, press again to confirm the preview. If you hold CTRL and press the mouse button instead, you will draw another preview, though it will not stay on screen if you release CTRL.

Variable-ind:
you under stood correctly. now the only thing left is to make the preview rectangles (or preview ellipse) follow the mouse till a mouse click (left for left tool and right for right tool) is pressed
exactly like what it does with the pencil tool here (except that we can drop the shapes ourselves instead of just a simple circle) 

https://user-images.githubusercontent.com/77773850/108486760-42817600-72c0-11eb-9a7f-0208aac9ac1f.mp4

KawanWeege:
> you under stood correctly. now the only thing left is to make the preview rectangles (or preview ellipse) follow the mouse till a mouse click (left for left tool and right for right tool) is pressed that would allow us to place the shape more accurately on the canvas
> 
> Exactly like what it does with the pencil tool here (except that for the shapes tool we can draw the shapes ourselves instead of just using a simple circle)
> 
>  pixelorama-2021-02-19_14.38.17.mp4

Ok, now the shapes should follow the mouse now.

Variable-ind:
Yep, works like a charm

Erevoid:
Extremely useful addition. I have some suggestions to make before it gets merged.

1. While holding shift the preview is a bit buggy, it keeps changing from where the cursor is and where the shape should be
2. The control button should be linked to the expand from center function, because alt+shift switches the language on windows machines, meaning it's not ideal for this purpose
3. It would be preferable if the whole idea with the current control button and the way the preview works is removed. We are considering of potentially making something similar with the shapes being selected after drawn with the 0.9 update and the selection reworks.
4. It would be nice if the tool options contained the word "size" to the left of the shape's size number and if it indicated the word "px" within the Spinbox, exactly like in the pencil options.
5. It would be even better if the ellipse and rectangle shapes are two different tools, instead of one tool containing both. 

Consider changing those elements and the PR is good to go!

KawanWeege:
> Extremely useful addition. I have some suggestions to make before it gets merged.
> 
> 1. While holding shift the preview is a bit buggy, it keeps changing from where the cursor is and where the shape should be
> 2. The control button should be linked to the expand from center function, because alt+shift switches the language on windows machines, meaning it's not ideal for this purpose
> 3. It would be preferable if the whole idea with the current control button and the way the preview works is removed. We are considering of potentially making something similar with the shapes being selected after drawn with the 0.9 update and the selection reworks.
> 4. It would be nice if the tool options contained the word "size" to the left of the shape's size number and if it indicated the word "px" within the Spinbox, exactly like in the pencil options.
> 5. It would be even better if the ellipse and rectangle shapes are two different tools, instead of one tool containing both.
> 
> Consider changing those elements and the PR is good to go!

About point 1, could you attach a video or image showing this problem? For me, it seems to be working correctly, the preview matches the drawn shape, and the way the shape scales while holding shift seems to be very similar to what you see in other pixel art software, like Piskel.
@Variable-ind asked for the implementation of the control button, so some discussion is needed to decide if we should remove it. I'll keep it for now, but I will change it to be the shift key instead.
The reason I added both shapes into one tool is to allow for the possibility of adding a variety of other shapes in the future, without polluting the toolbar.

Variable-ind:
how about this, we could remove the control instead we could set it in such a way that a single click will move the preview where the mouse is currently at and double-click will draw the shape (or an "Apply" button can simply be placed in the tool settings instead of double-click)

OverloadedOrama:
> About point 1, could you attach a video or image showing this problem? For me, it seems to be working correctly, the preview matches the drawn shape, and the way the shape scales while holding shift and moving the mouse seems to be very similar to what you see in other pixel art software, like Piskel.

I can replicate the issue on Windows 10.

https://user-images.githubusercontent.com/35376950/108897395-e756c880-75ca-11eb-90ea-8f3008b50627.mp4

In the video I am always holding the Shift button, and the end result is correct, but the preview is flickering.

OverloadedOrama:
As for the other points, I'd say to remove the Control functionality completely (and use Control for the behavior that is used in Alt) for now and discuss it again in the future once the new selection system has been finalized, in case we can use it to move the shape after it has been created. So, Shift for keeping aspect ratio, Control for expanding from center.

I also think putting the 2 shapes as 2 different tools would be better, it would allow easier switching for the users as they can map them to different keyboard shortcuts. New tools will be added in v0.9 so the pollution problem will most likely occur anyway, but that's a different issue. We could group tools together in a single icon, like GIMP and Aseprite does it, if it becomes a problem.

KawanWeege:
Alright, I think I've addressed most of these issues. I've made various changes so more testing is needed.


> 1. While holding shift the preview is a bit buggy, it keeps changing from where the cursor is and where the shape should be

This is supposed to be fixed now

> 2. The control button should be linked to the expand from center function, because alt+shift switches the language on windows machines, meaning it's not ideal for this purpose
> 3. It would be preferable if the whole idea with the current control button and the way the preview works is removed. We are considering of potentially making something similar with the shapes being selected after drawn with the 0.9 update and the selection reworks.

The previous implementation of the ctrl button has been removed. Now, ctrl is used to draw 1:1 shapes.

> 4. It would be nice if the tool options contained the word "size" to the left of the shape's size number and if it indicated the word "px" within the Spinbox, exactly like in the pencil options.

Added those as well

> 5. It would be even better if the ellipse and rectangle shapes are two different tools, instead of one tool containing both.

Made them separate

DragonOfWar:
Should I change the expand from click origin button to be CTRL instead of ALT and make the draw 1:1 shape button be SHIFT, or the current configuration (CTRL = Draw 1:1 / ALT = Expand from click origin) is ok?

Variable-ind:
You should change the alt button to ctrl.

OverloadedOrama:
Yeah I think Ctrl for expand from center and Shift for 1:1 might be better. In the future we'll make these shortcuts customizable for the users anyway, so everyone can have their own configuration. (#309)

DragonOfWar:
> Yeah I think Ctrl for expand from center and Shift for 1:1 might be better. In the future we'll make these shortcuts customizable for the users anyway, so everyone can have their own configuration. (#309)

Done. If I haven't missed anything, it should be good to go.

Erevoid:
Looking nice, I too think it's good to go

PR #458, opened 2021-02-22
full support for auto tallscreen/widescreen panel layout

ballerburg9005:
If the window aspect ratio is lowered below 3:4, then the panel layout will switch to "tallscreen", which entails that the right toolbar will be moved to the bottom instead. Of course the reverse is also true. The user can manually override the panel layout behavior in View -> Panel Layout.

Tested with Godot 3.2.1 on Linux and HTML5 with Chrome. Should be pretty error-proof though.

tallscreen_mode2

This illustrates the maximum usable drawing area with the different modes:

tallscreen_vs_widescreen


ballerburg9005:
The canvas preview now hides, when there is not enough room due to display scaling.

tallscreen_mode2

ballerburg9005:
The canvas preview now hides, when there is not enough room due to display scaling.

tallscreen_mode2

ballerburg9005:
tallscreen_vs_widescreen

Erevoid:
Pixelorama supporting vertical screens is a very important addition and thanks for making it. I made a concept of a more organised version of what you made, with a bit more drawing space and a slight visual UI placement improvement. What do you think?
suggestion

ballerburg9005:
You are welcome! I was afraid that you maybe wouldn't appreciate tallscreen support.

I have made the changes, except for the color picker, which I overlooked.

I am a little bit hesitant to move too many elements around, as later changes in the UI would also require the script to change.

Moving the color picker is a bit ... hmm, I really don't know!

Check out how it is so far:

tallscreen

I later noticed that it is most important to minimize the mouse distance to the tools. With the tools tool panel at the bottom, it is a tad bit longer, but you will make the same downward movement every time, which I believe is more intuitive.

If you still really want the color picker above the CanvasPreview, I can do it as well.

Erevoid:
I'm a fan of the layers being moved on the bottom. It looks good so far! The tools may need a second column in the future and that would greatly decrease the canvas area visible in vertical screens. While the colour picker looks good when the preview is removed, when the preview is there it has empty space under it and it seems unbalanced (visually). I would suggest having it at its current position without the preview and when the preview is enabled have it over it? As for the tools being moved down there, it may even require less mouse movements if you need to choose a tool and change its tool options so I think it's ideal now!

ballerburg9005:
Final result. Please merge.

ts2

src/Main.gd
```diff
@@ -4,6 +4,16 @@ var opensprite_file_selected := false
...
+var tallscreen_is_active = false
+onready var UI                     := $MenuAndUI/UI
+onready	var BottomPanel            := $MenuAndUI/UI/CanvasAndTimeline/HBoxContainer/BottomPanel
+onready var RightPanel             := $MenuAndUI/UI/RightPanel
+onready var ToolAndPaletteVSplit   := $MenuAndUI/UI/RightPanel/PreviewAndPalettes/ToolAndPaletteVSplit
+onready var ColorAndToolOptions    := $MenuAndUI/UI/RightPanel/PreviewAndPalettes/ToolAndPaletteVSplit/ColorAndToolOptions
+onready var CanvasPreviewContainer := $MenuAndUI/UI/RightPanel/PreviewAndPalettes/CanvasPreviewContainer
+onready var ToolPanel              := $MenuAndUI/UI/ToolPanel
+onready var myScrollContainer   := $MenuAndUI/UI/RightPanel/PreviewAndPalettes/ToolAndPaletteVSplit/ColorAndToolOptions/ScrollContainer
+
```
OverloadedOrama, line 16: Minor issue, but for readability purposes, make sure that the code follows the GDScript style guide, namely the [whitespace guide](https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_styleguide.html#whitespace) and [variable naming conventions](https://docs.godotengine.org/en/stable/getting_started/scripting/gdscript/gdscript_styleguide.html#functions-and-variables). Variable names should be in snake_case.

OverloadedOrama (changes requested):

ballerburg9005:
It's your program, but I would suggest you keep it as it was. I specifically made exceptions to the rule, because it turned out to be more readable and the variable names 1:1 matched the actual Node names, which is cleaner and more useful.

src/Main.gd
ballerburg9005: It's your program, but I would suggest you keep it as it was. I specifically made exceptions to the rule, because it turned out to be more readable and the variable names 1:1 matched the actual Node names, which is cleaner and more useful.

OverloadedOrama: I understand the reasoning, but since the rest of the code follows (or at least, tries to follow) the GDScript style guide, for the sake of code consistency, I think it would be better if there are no exceptions. I plan to go through all of the rest of the code at some point and spot inconsistencies that may have occurred and fix them.

Alternatively, we could rewrite all of the other node variable names to match the node names (or rename the nodes to match the variables), but I'm not sure if this is worth the effort.

src/Classes/Project.gd
```diff
@@ -208,6 +209,8 @@ func change_project() -> void:
 
 	for j in Global.TileMode.values():
 		Global.tile_mode_submenu.set_item_checked(j, j == tile_mode)
+	for j in Global.PanelLayout.values():
+		Global.panel_layout_submenu.set_item_checked(j, j == panel_layout)
```
OverloadedOrama, line 213: Is the panel layout meant to be a per-project setting? Because if I, for example, select Tallscreen and then switch tabs, the submenu item that is checked returns to Auto, but the UI does not change.

Per-project setting means that the UI would change depending on the project, each project would have its own panel layout. If this is something we want, then this code also needs to call `change_ui_layout()` from Main.gd. If this is something we don't want, then these 3 new lines of code should be removed completely. Personally, I don't think it should be a per-project setting as it's a UI change that may confuse the users.

src/UI/TopMenuContainer.gd
```diff
@@ -314,6 +323,13 @@ func window_transparency_submenu_id_pressed(id : float) -> void:
 		window_transparency(id/10)
 
 
+func panel_layout_submenu_id_pressed(id : int) -> void:
+	Global.current_project.panel_layout = id
+	for i in Global.TileMode.values():
```
OverloadedOrama, line 328: I think this line should be `for i in Global.PanelLayout.values():`.

OverloadedOrama (commented):
Tested it more and it seems to be working well. Left a couple more comments and after they get resolved, it should be ready to get merged!

src/Classes/Project.gd
ballerburg9005: It will now save and restore the layout if you close and reopen the window. 

On that matter: shouldn't most other global view settings also save and restore? If the users sets a different window transparency or hides the animation timeline for example. Maybe not tile mode though and especially not mirror view.

Honestly, when I tied panel_layout to current_project I wasn't sure whether or not saving the view settings were broken. So I didn't think it through any further.

src/UI/TopMenuContainer.gd
ballerburg9005: Check out new commit.

ballerburg9005 (commented):
Check out new commit.

src/UI/TopMenuContainer.gd
ballerburg9005: fixed

OverloadedOrama (approved):
Looks good to me. Thank you!

PR #459, opened 2021-02-23
Add shape fill tool

EIREXE:
Adds a tool that lets the user draw a shape that is filled once they let go.

Video: https://youtu.be/tnci7R-fiTQ


Erevoid:
I like this addition. Can it be implemented similarly to how the brush tool works? Either as part of the brush tool, enabling it from the brush tool options or like now, having its own tool but with the same options the brush tool gets (brush size, different brushes, pixel perfect etc)

novhack:
I would also like it more as a property for a Pencil tool rather than a separate tool.

Variable-ind:
i did it in https://github.com/Orama-Interactive/Pixelorama/pull/489 feel free to take a look. :heart:

OverloadedOrama:
Superseded by #489. Thank you for your PR, since #489 is based on this, we will of course credit you too!

PR #473, opened 2021-04-01
Fix hue and saturation getting reset when draging value slider to zero

kevinms:
The color picker widget in HSV mode was resetting HS when you drag V to zero.

Thought about changing Tools.assign_color() to take an optional `signal_color_changed` boolean. But, I can imagine other objects connected to `Tool.color_changed` that may want dynamic updates as you use a color picker in the future. So the fix here is to temporarily disconnect `ColorAndToolOptions` from `Tool.color_changed` while the popup is active. This isolates the fix to the thing that needs fixing.

### Bug:
hsv-slider-bug

### Fix:
hsv-slider-fix


OverloadedOrama (approved):
Looks good. Thanks!

PR #490, opened 2021-05-24
Set project name

Variable-ind:
gave the ability to set project name on startup,
And also fixed a bug that "_if the aspect ratio button is pressed and a template of different width and height were selected, then the button also applied on them e.g a template of (but it shouldn't)_"
**Demonstration of bug**: write 1 in width, 2 in height, enable aspect ratio, and then try to change template... a template of 256x240 will give 120x240


OverloadedOrama (approved):
Awesome, thanks!

PR #494, opened 2021-06-14
Use CMD+Shift+H for showing the pixel grid on OSX

### LINKED ISSUE #493: macOS show pixel grid hotkey conflicts with system hotkey
**Pixelorama version:**

v0.8.3-stable

macOS 10.14.16

**Issue description:**

On macOS command + h is a system wide shortcut to hide the current window, so it can't be used to toggle show the pixel grid

**Steps to reproduce:**

1. Open Pixelorama on a Mac
2. Press command + h
3. Window will hide and pixel grid will remain in the current state when it is reopened

I don't really have a suggested fix for this (nor am I bothered by it). If you close as won't fix I won't be offended, just figured I would point it out in case you weren't aware.

Schweini07:
Fixes #493

It's debatable if it's the right choice, just using Shift, but I think it's fine.

### DIFF
--- a/src/Autoload/Global.gd
+++ b/src/Autoload/Global.gd
@@ -678,6 +678,9 @@ func use_osx_shortcuts() -> void:
 
 	for action in inputmap.get_actions():
 		var event : InputEvent = inputmap.get_action_list(action)[0]
+		
+		if event.is_action("show_pixel_grid"):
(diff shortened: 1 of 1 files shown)


OverloadedOrama (approved):
Thanks!

Issue #343, opened 2020-09-30
Use shaders instead of Gdscript for HSV adjustments


anasofgo:
I noticed that the performance drops significantly when you have a large picture you are working on. I tried to make a 1000x1000 art work (which is not that big) and if I try filling it with color the program freezes for a few seconds.. I then try adjusting the HSV from the "Image" menu and the program crashes.

I tried to dig up the issue and I found the following function:

```
func adjust_hsv(img: Image, delta_h : float, delta_s : float, delta_v : float, pixels : Array) -> void:
	img.lock()
	for i in pixels:
		var c : Color = img.get_pixelv(i)
		# Hue
		var hue = range_lerp(c.h,0,1,-180,180)
		hue = hue + delta_h

		while(hue >= 180):
			hue -= 360
		while(hue < -180):
			hue += 360

		# Saturation
		var sat = c.s
		if delta_s > 0:
			sat = range_lerp(delta_s,0,100,c.s,1)
		elif delta_s < 0:
			sat = range_lerp(delta_s,-100,0,0,c.s)

		# Value
		var val = c.v
		if delta_v > 0:
			val = range_lerp(delta_v,0,100,c.v,1)
		elif delta_v < 0:
			val = range_lerp(delta_v,-100,0,0,c.v)

		c.h = range_lerp(hue,-180,180,0,1)
		c.s = sat
		c.v = val
		img.set_pixelv(i,c)

	img.unlock()

```
This for loop is apparently very taxing in terms of computational power, and it is also used for the "preview" window.

If there is no more efficient workaround solution for this, I think it would be wise to make it be called out only when the user confirms his adjustments. it shouldn't be used to update the little "preview" window.

For the preview, I suggest that a simple hsv adjustment shader is used. Shaders are faster and more efficient for "realtime" previews.

Another additional thing that I found interesting is this https://github.com/Coldragon/godot-shader-to-image . Essentially a shader can be turned into an image, so it is possible to add extra functionality and effects more easily using shaders. I am still experimenting with this approach.

The following are my Laptop specs:
Nvidia Gtx 1060
intel Core i7
16 GB ram
Windows 10

luiq54:
HSV adjustment dialog has been moved to shader based one in a [PR](https://github.com/Orama-Interactive/Pixelorama/pull/475). Can you try to get a build from [here](https://github.com/Orama-Interactive/Pixelorama/actions/runs/767494430) and compare results ?

anasofgo:
[...]

PR #510, opened 2021-07-18
Faster Preview for HSV

luiq54:
1) Added checkbox in HSV Dialog to shrink preview image by 3 times as mentioned in #343.
2) Disables filter to make previews clearer 
3) Added `shrink_preview` to `ImageEffects.gd` so that any dialog can hook into it.

### DIFF
--- a/src/Classes/ImageEffect.gd
+++ b/src/Classes/ImageEffect.gd
@@ -13,6 +13,7 @@ var preview_texture : ImageTexture
 var preview : TextureRect
 var selection_checkbox : CheckBox
 var affect_option_button : OptionButton
+var shrink_preview: bool = false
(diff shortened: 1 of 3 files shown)


Variable-ind:
Just tried it and it seems to work fine... (but you forgot to turn off visibility in "HSVDialog.tscn")

luiq54:
> 
> 
> but you forgot to turn off visibility of "HSVDialog") 
Oof , my bad. Yeah I will fix it!

luiq54:
I am closing this request as I wasn't able to refactor the code into something understandable and current implementation has performance tradeoffs which are not worth the gains. I will attempt to refactor it again in future.

PR #514, opened 2021-07-30
Tweak UI to use less empty space

novhack:
These are few little changes to make the right panel a bit more compact with smaller margins.
Before:
Screenshot_20210730_235850
After:
Screenshot_20210730_235541


novhack:
Added a fake grabber to tools panel. It removes the border next to it.
Tweaked some margins in animation panel.
Fixed a few issues with resizing.
Resized the preview by 10px. Some space was saved by moving swap/default color buttons.
Tallscreen mode is a bit broken now. I will fix it in the next commit.
Screenshot_20210801_221131

Variable-ind:
Works fine. I like it!

novhack:
I forgot to mention that I think that I resolved all issues with tallscreen mode so it should be ready to merge if everybody likes it.

OverloadedOrama (changes requested):
I think it looks good! One small problem though, when opening up the Preferences window, the Interface options are visible from the Startup tab. You may have set them to be visible by default. Other than that, it should be ready to get merged.

novhack:
Oh, sorry! If you want to fix that you can otherwise I will fix it in a few hours tonight.

OverloadedOrama (approved):
Should be all good now. Thank you!

PR #515, opened 2021-07-31
Update pt_PT.po

Valeriooooh:
Full portuguese translation and localization. :)


novhack:
Thank you! Pixelorama uses Crowdin (https://crowdin.com/project/pixelorama) to manage translations and this file is generated.
I'm not sure if it can be merged because it would get probably overwritten when translations are merged so it probably must be imported to Crowdin.

Valeriooooh:
I've imported the file and finished translating the rest. :) hope it is all good.

PR #519, opened 2021-08-18
Fixed issues with Shading tool Saturation and Value not always being right in Hue Shading mode

mrtripie:
This should fix some problems with the Saturation and Value part of the Shading Tool (NOT the Hue problems).

image
Left grey theme is previous build, right side is new one.

Problem 1: If you darkening value is more than the value of the color, it can end up brightening instead (I think Godot's Color class might wrap around values when setting them)
image
(Checked the color with the color picker to make sure it was right hue and at the value limit of 10)

Problem 2: If you darken not quite enough for that, but still too much, it will hit the value limit and ignore the darkening. Now it will just be set to the Value limit:
image

Problem 3: When lightening too much using Saturation, the hue could get messed up, I'm not entirely sure why this happened, but applying the same fix I did to the value part seemed it:
image

Problem 4: If you brightened the Saturation too much, it would also through out the Hue and Value brightening, so I erased that part. Its not really obvious between the old and new version, so for these screenshots I added it back:
Before: (No value change as it increased the saturation all the way)
image

After: Both Saturation goes down and Value go up, making it White.
image

Not sure if  it should need more HSV brightening/darkening limits, but it should feel much better to use now.

### DIFF
--- a/src/Tools/Shading.gd
+++ b/src/Tools/Shading.gd
@@ -56,20 +56,16 @@ class LightenDarkenOp extends Drawer.ColorOp:
 			if lighten_or_darken == LightenDarken.LIGHTEN:
 				hue_shift = hue_limit_lighten(dst.h, hue_shift)
 				dst.h += hue_shift
-				dst.s -= sat_shift
+				if dst.s > sat_lighten_limit:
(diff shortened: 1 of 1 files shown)


OverloadedOrama (approved):
Looks good, thank you!

PR #522, opened 2021-08-23
Shading Tool Hue Shift Fixes + Tweaks

mrtripie:
- Changed hue_darken_limit from purple to blue.

- Limited the Hue Shift slider to a range of 180 so that half of it wouldn't always hit the hue shift limits.

- Fixed hue shifting issues:
  - Lightening past the limit would do nothing for some colors rather than set it to the hue limit: (Left side before, right side after)
image
  - Same happened with darkening for some colors (the red and magenta patches), while others would end up not being limited at all, instead going to incorrect hues (everywhere should have been changed purple).
image

- Also:

- When finally shifting the hue used fposmod to work around a Godot Color bug that made shifting the hue around the "ends" of the color wheel not always return the correct hue.

- Replaced 359.0 in code to 360.0 which fixed some instances of the hue shift being 1 off the expected value.

- To make a little easier to read removed the **!** on the hue_range check and flipped the results of the hue_range function.

- Updated comments to reflect changes, and make color ranges a little more specific.

- NOTE: Added a PR in the documentation repository for changing the shading limit from purple to blue.

### DIFF
--- a/src/Tools/Shading.gd
+++ b/src/Tools/Shading.gd
@@ -23,8 +23,8 @@ class LightenDarkenOp extends Drawer.ColorOp:
 	var sat_amount := 10.0
 	var value_amount := 10.0
 
-	var hue_lighten_limit := 60.0 / 359.0 # A yellow color
-	var hue_darken_limit := 270.0 / 359.0 # A purple color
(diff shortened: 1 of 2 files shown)


OverloadedOrama (approved):
Looks great, thank you!

PR #526, opened 2021-08-29
DrawingAlgos Explicit Project DataType

behreajj:
I needed to add the data type to project for the scripts to compile. Pardon any mistakes, I'm still working through contributing guidelines.


OverloadedOrama (approved):
I wonder why that happens. I suppose it's Godot bug, I have noticed similar behavior happening to situations like these when I open forks of Pixelorama in Godot or from another device. Anyway, thanks!

PR #527, opened 2021-08-30
Selection fixes

Variable-ind:
Some selection fixes regarding:
loss of data on toggling layer buttons,
and moving content of multiple selected cels when one or more cels are in a locked or invisible layer


OverloadedOrama (approved):
Thank you!

PR #531, opened 2021-09-03
Boost HSV and Rotation sliders

Variable-ind:
It will hopefully solve #343
**Working Principle:**
It uses a timer with a certain wait_time (10 mili-second by default), If no more values are changed during that interval then it updates the preview, It will treat the rapidly changing values as useless and will skip those values. As a result it can potentially improve performance significantly
**Testing:**
By default it's set to 10mili-seconds because the user normally draws on a small (say 64x64)
canvas, But if you want to change HSV or Rotation for say a large (say 1024x1024)canvas you can set Wait (in miliseconds) to see the changes on preview to a high value (e.g 100msec is ideal for a canvas that size) and it will start filtering out useless values (Try it out and tell me what you think about it :smile:)

**A demonstration of it's power**

https://user-images.githubusercontent.com/77773850/131991396-761e0e83-5b13-4e77-836c-38e8374d8633.mp4


OverloadedOrama (approved):
Looks good, thank you!

PR #538, opened 2021-09-19
Add buttons for looking at different Splashscreen Artworks

Schweini07:
image
Two arrows are now located at the sides of the current artwork. When pressed the artwork changes to the one next in the array of artworks. The array was a dictionary before, but using it simplifies everything a lot.
Everything should be accounted for, the arrow icons change according to theme and the code checks that the array doesn't go out of bounds.

### DIFF
--- a/src/UI/Dialogs/SplashDialog.gd
+++ b/src/UI/Dialogs/SplashDialog.gd
@@ -1,21 +1,24 @@
 extends WindowDialog
 
 
-var artworks := {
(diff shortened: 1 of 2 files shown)


Schweini07:
Had to make a minor change, because I overlooked something. Commits are rebased and everything should be fine👍.

OverloadedOrama (approved):
Looks good, thank you!

PR #543, opened 2021-09-28
Add idle fps setting

Schweini07:
A new setting was added trough which the user can set how many FPS Pixelorama should have when idle.
image
Additionally a bug was fixed where when the user has another application as their focus and reenters Pixelorama with the mouse but not focusing it and then exiting with the mouse, the target_fps would be set to the standard.

### DIFF
--- a/Translations/Translations.pot
+++ b/Translations/Translations.pot
@@ -1055,13 +1055,13 @@ msgstr ""
 msgid "Set application FPS limit:"
 msgstr ""
 
-msgid "Limit FPS to 1 when app loses focus"
+msgid "Limit FPS to the idle FPS when app loses focus"
 msgstr ""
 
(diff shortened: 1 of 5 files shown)


Variable-ind:
Also i think the configuration dialog should be hidden by default and only appear when the checkbox is enabled (cause it's associated with the checkbox)

Schweini07:
Good catch, I forgot about that!
Regarding your suggestion to hide the setting, I think that should not be done, it hides useful information from the user and includes extra work that may not be worth it.

Variable-ind:
But the field is only useful when the checkbox is checked, i don't think it serves a purpose if the checkbox is unchecked. i mean something like what i did with live preview checkbox in HSV dialogue.

OverloadedOrama (approved):
Looks good, thank you!

(In case someone was not present in our discussion in Discord, we decided to disable the "Set idle FPS" setting when Limit FPS is off, instead of hiding it)

Issue #501, opened 2021-06-23
Lag  when using 20px brush on a 512x512 canvas caused many calls inside draw_fill_gap 


Nagamine-j:
**Pixelorama version:**
5b74a4292d874bc69c76adf8e3c7ddbbb8b4dd2e

**OS/device including version:**
Windows 10 / laptop ryzen 5 amd vega 8

**Issue description:**

Lag when using 20px brush on a 512x512 canvas, It's cause because the 50 000 calls to can pixel get wich are caused by the draw_fill_gap function
![image](https://user-images.githubusercontent.com/60224203/123057243-51db1200-d3cd-11eb-826f-bcf6289b09e0.png)
https://github.com/Orama-Interactive/Pixelorama/blob/5b74a4292d874bc69c76adf8e3c7ddbbb8b4dd2e/src/Tools/Draw.gd#L202-L221

looking at the algorithm on  https://godotengine.org/qa/35276/tile-based-line-drawing-algorithm-efficiency
I think the draw_tool(Vector2(x, y))  function call should be outside the while

**Steps to reproduce:**
1. Create a 512 x512 project
2. select 20px brush
3. draw

Nagamine-j:
Ok i checked again and the draw_tool function call goes inside the while, but it calls draw_tool incrementing the diferential pixel by pixel and every draw_pixel calls set_pixel 400 times (which is understandable because the brush is 20*20) however many of this function calls are unncesary because most of the pixels are already set in the previews draw_pixel call.
I think it should be more optimized approach

Docmine17:
I have the same problem here, I use Ubuntu 21.04.

Variable-ind:
Hy :wave:, i tried to optimize it a bit [here's](https://nightly.link/Orama-Interactive/Pixelorama/actions/runs/1918042931) the nightly link to it. Does this solve the problem?

MatteoPiovanelli-Laser:
Hello.
I did some tests on this. The `_cache_size` as it's being computed seems to be helpful when the brush type is "Pixel brush". For other types its usefulness in making things faster varies. I changed the `_set_pixel(position: Vector2, ignore_mirroring := false)` method as follows:

```
func _set_pixel(position: Vector2, ignore_mirroring := false) -> void:
	if position in _draw_cache and _for_frame == Global.current_project.current_frame:
		return
	_update_brush_cache(position)

	var project: Project = Global.current_project
[...]

PR #554, opened 2021-10-28
Optimize the drawing of large amounts of pixels

sayaks:
#501 complains of lag when drawing a lot of pixels at once, a major reason for this lag seems to be a lot of redundant drawing attempts. Using GDNative here or C# will probably help, but a lot can be gained by just optimizing the algorithms used.

Doing this right though will require me to take some time to understand the entire flow of how pixels are drawn, so I'm making this a draft pull request until i can be more sure that what I've done is correct.


OverloadedOrama:
Superseded by #657.

PR #557, opened 2021-11-08
Use Luminance In Desaturate Shader

behreajj:
Updated desaturate shader to use perceptual luminance instead of HSV. See https://en.wikipedia.org/wiki/HSL_and_HSV#Disadvantages

Before:

before

After:

after

Thank you for considering this pull request.
Best,
Jeremy


OverloadedOrama (approved):
Looks good to me. Maybe in the future we could let the user choose a desaturation method like other software do, but I think this does the job for now so I'll go ahead and merge the changes. Thank you!

PR #558, opened 2021-11-11
Canvas Rotation

mrtripie:
Implemented Canvas Rotation.

Used @Variable-ind 's work as a base for this, where the viewport camera is rotated to make the canvas rotate.

-Changed the zoom label on the top bar (that could be double clicked to be edited) to a button which will be highlighted when hovered over and gave it the pointing finger mouse cursor to make it more discoverable that it can be pressed and edited. It also now only needs a single click to edit.
-Added a matching rotation button that can also be used to edit the rotation.
    - Note that the value shown on this is multiplied by -1 to make going up rotate clockwise, which feels more natural. (This is because rotating the camera in one direction makes the canvas rotate in the opposite direction from the camera's perspective)
-Fixed several little things that didn't work quite right with rotation such as as zooming to the mouse location.
-Made moving the selection with the arrow keys move with the canvas rotation considered (snapped to 90 degrees)
-Made fit to frame take the rotation into consideration
-Gave each project their own rotation variables so that switching between projects restores their rotation (the same as with offset and zoom)
- Adjust the camera offset while rotating to make it so that it rotates relative to the center of the project (not relative to the camera offset)
- Adjust the ruler positions so that the horizontal ruler's zero position matches with the farthest left corner of the project, and the vertical ruler's zero position matches with the top corner of the project.
- Made it so when dragging guides out of rulers, it takes into consideration the camera rotation when setting if the created guide is a Horizontal or Vertical one.
- Made it so that the text showing the guide location when moving guides position stays on screen even when rotated by using Godot's Geometry class to test which viewport edges the guide intersects with. If its outside of the viewport it just puts the text in the top left corner.


Variable-ind:
It works great!, by the way i can't seem to find the buttons (that were at the top-right) for zoom and drag... are they removed?

OverloadedOrama (approved):
This looks great, excellent job! I tested it and I didn't encounter any issues, so it should be good to go.

Thank you!

@Variable-ind The buttons were moved to a separate branch, so I assume there will be another PR for that.

mrtripie:
@Variable-ind, @OverloadedOrama Thanks!

@Variable-ind Yeah, I figured that it would make sense for them to be separated into separate branches and PRs. As the rotation button relied on canvas rotation being implemented, I finished this up first. Next, I'll work on polishing up the buttons.

PR #559, opened 2021-11-14
Cache Save/Open Sprite Dialog's directory, and keep dialogs synced

mrtripie:
- Added to the cache file a "file_dialog_dir" value that can be used to set the Save/Open Sprite Dialog's current_dir when opening Pixelorama, as it can be a bit annoying to have to navigate to your game's project folder (or general art folder) each time you open Pixelorama.
- Whenever you open a file from the Open Dialog or save a file from the Save Dialog, it sets the current_dir of the other, and sets the cache value.
- When using "Open Last Project" or "Recent Projects" it will also set the current_dir of both Open and Save Dialogs, as well as sets the cache value.

- For exporting files:
  - Exporting doesn't set the file dialog path's or caches the directory, as the user may want to keep their project and export files in separate places
  - For exporting a file that doesn't already have a path, use the cached directory over the Desktop directory when possible:
    - If Export Dialog's directory_path is empty, use the cached directory (or Desktop if none)
    - In some situations an untitled project would still bring up the Desktop directory when exporting, searching the code for the Desktop directory I found in Project the directory_path property, which setting that to the cached directory in init over Desktop when possible seems to have fixed this.


mrtripie:
Maybe it might make sense to have a separate cached "export_dir" value for exporting files that don't have a saved project, rather than reusing the "file_dialog_dir" value?

mrtripie:
Maybe "current_dir" would be a better name than "file_dialog_dir"?

OverloadedOrama:
> Maybe it might make sense to have a separate cached "export_dir" value for exporting files that don't have a saved project, rather than reusing the "file_dialog_dir" value, as some users will want to keep their projects and exported files seperate?

Hmm, I'm not sure what would be better. I like the current behavior of the PR. Unless there are disagreements, maybe we can keep what you made and change it in the future if users want to.

> Maybe "current_dir" would be a better name than "file_dialog_dir"?

Yes I think it would be a better name.

mrtripie:
@OverloadedOrama  Alright, changed the name to "current_dir".

OverloadedOrama (approved):
Awesome, thank you!

PR #560, opened 2021-11-15
Improvements to Animation Tags

Variable-ind:
The PR contains;

1. Fixed the offset issue (significantly noticeable on tags larger than 20)
2. Fixed upper line of tag vanishing on greater window scales
3. Added a randomizer to autoselect colors of new tags
4. Fixed Tags scrolling without their cels when scrolled on them

**Before:**
before

**After:**
after


src/UI/Timeline/FrameTagDialog.gd
```diff
@@ -65,6 +65,8 @@ func _on_AddTag_pressed() -> void:
 	current_tag_id = Global.current_project.animation_tags.size()
 	options_dialog.get_node("GridContainer/FromSpinBox").value = Global.current_project.current_frame + 1
 	options_dialog.get_node("GridContainer/ToSpinBox").value = Global.current_project.current_frame + 1
+	randomize()
```
OverloadedOrama, line 68: `randomize()` is not needed here. It's already called in the first line of Global's `_ready()` method and [it should only be called once at the start of the project](https://docs.godotengine.org/en/stable/tutorials/math/random_number_generation.html#the-randomize-method).

Variable-ind: Oh,  i didn't know!. Okay, i removed it.

OverloadedOrama (approved):
Looks good, thank you!

PR #561, opened 2021-11-16
Guide Color Improvements

mrtripie:
- Use the Label font_color from the current theme to set the color of the position text on guides (allowing it to be seen more easily on the light theme)
- Made Symmetry guide use default_color instead of modulate for its color, which makes the position text color correct.
- Gave back Symmetry Guide's slight color variation from regular guides by mixing in some semi-dark, semi-saturated blue. This gives it a similar color to how it was before with the default preference, and works okay with most colors.
- Made guides that are outside the canvas (and will be freed if released) semi-transparent (doesn't happen with symmetry guides)
- Moved the side effect of freeing the guide outside of the outside_canvas function (and into input) to allow to use it elsewhere
- Fixed symmetry guides not having their color updated correctly when changing the preference


OverloadedOrama (approved):
Looks great, thank you!

PR #598, opened 2021-11-18
Raspberry 4 build

OverloadedOrama:
A GitHub Actions workflow that generates a Raspberry 4 Pi build for Pixelorama, using the [Unofficial Godot Engine Raspberry Pi binaries](https://github.com/hiulit/Unofficial-Godot-Engine-Raspberry-Pi), as suggested by @ragreenburg. ARM builds were requested in #594, but I am not entirely sure whether these binaries run on all 32-bit ARM systems, or just Raspberry Pi 4. Maybe using [FRT](https://github.com/efornara/frt) would be better, but seeing as the last commit was back in April, it doesn't look like it gets updated often, so it might not be a good idea.

I don't have a RPI 4 or any Linux ARM device to test it myself, so testing from someone else is needed. To try it, simply head to the "Checks" tab in this Pull Request, on the left side click on "dev-desktop-builds" and download the Linux-RPI4 artifact. Unzip the file, and run "Pixelorama.rpi4".


ragreenburg:
Sorry, I should have been quicker to test/respond. I tried building on my Pine Phone (Linux ARM) and that didn't work though I will say I don't think it matter since the screen doesn't fit the phone and it would take a lot of work to make that happen so I don't expect a phone build. I have reached out to someone who has a raspberry pi 400 to test this build though. I will let you know when I get a response! Thank you though for working on this!

ragreenburg:
I got confirmation that it launches on a Raspberry Pi 4!

OverloadedOrama:
That's awesome! Thank you for your help!

PR #600, opened 2021-11-25
Better Onion Skinning

Variable-ind:
**The main goal of this PR is to expose more options for onion so that it can adapt to any situation and user need**

https://user-images.githubusercontent.com/77773850/118992553-2fa13f00-b99e-11eb-8d95-88109d6034bd.mp4


OverloadedOrama (approved):
Nice, thank you!

PR #603, opened 2021-12-05
Fix drawing bug reported in issue-555

### LINKED ISSUE #555: Undoing while in the middle of drawing causes unexpected behavior
**Pixelorama version:** 0.9

**OS/device including version:** Linux 5.10.70-1-MANJARO

**Issue description:** Undoing while in the middle of drawing seems to cause previously undone things to reappear.

**Steps to reproduce:**
1. Draw line with pencil
2. Draw second line with pencil, and while holding the mouse button down undo. Now line 1 and 2 should disappear.
3. Draw third line with pencil, and while holding the mouse button down undo . Now line 1 should reappear, and line 3 should disappear.

I would think that what should happen is that only the lines drawn in step 2 and 3 should disappear, but never the line drawn in step 1.

JumpJetAvocado:
I have fixed a bug reported in [https://github.com/Orama-Interactive/Pixelorama/issues/555](url) which happens when undo is performed while drawing with pencil/eraser/shading tools.

undo_1
The gif above is a bug I reproduced by following the steps described in the issue. The first line reappears if undo is performed while drawing the third line. But it can be removed by performing undo once more.

undo_2
This one would be the worst case scenario. If undo is performed twice after drawing the third line, The reappeared first line cannot be removed with undo so it should be removed manually.

The simplest solution I have come up with was moving `project.undo_redo.create_action(action)` in `commit_undo()` in Draw.gd to `prepare_undo()` so undo would be disabled while drawing. Due to this change, I also have modified `func prepare_undo() -> void:` to take 1 String parameter like `func prepare_undo(action: String) -> void:` and some scripts calling this method.

undo_3
One side effect is that Line/Shape tools are also affected by this change. So actions like above will no longer be possible. Though it might be considered as a nice change since actions like that could have potentially made redo difficult by overwriting the undo list with a new change.

Maintainer edit: Closes #555

### DIFF
--- a/src/Tools/Draw.gd
+++ b/src/Tools/Draw.gd
@@ -152,7 +152,7 @@ func prepare_undo(action: String) -> void:
 	project.undo_redo.create_action(action)
 
 
-func commit_undo(action: String) -> void:
+func commit_undo() -> void:
 	var redo_data := _get_undo_data()
 	var project: Project = Global.current_project
 	var frame := -1
--- a/src/Tools/Draw.gd
+++ b/src/Tools/Draw.gd
@@ -146,8 +146,10 @@ func update_line_polylines(start: Vector2, end: Vector2) -> void:
 	_line_polylines = _create_polylines(indicator)
 
 
-func prepare_undo() -> void:
+func prepare_undo(action: String) -> void:
+	var project: Project = Global.current_project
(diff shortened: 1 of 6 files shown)


OverloadedOrama (approved):
It seems to be working well. Thank you!

PR #623, opened 2021-12-25
Spritesheet layer improvements

Variable-ind:
1. Drastically changed the "Spritesheet layer" function and removed its dependence on other functions, hence reducing the amount of undos (i hope didn't break anything though)
2. "Start frame" can now be as high as you want.
3. a minor UI fix regarding "PreviewDialog.tscn"


OverloadedOrama (approved):
Looks good, thank you!

PR #625, opened 2021-12-30
Shader alpha fix

### LINKED ISSUE #615: Effects change pixels outside of selection
**Pixelorama version:**
Pixelorama master

Arch, AMD RX5700 XT - mesa 21.2.5

**Issue description:**
It seems that some Image shader effects like desaturation affect pixels with alpha<255 outside of selection even if "Only affect selection" is toggled. 
obrazek
I think that the problem is somewhere in `ShaderImageEffect.gd` because I encountered it also while trying to use it for bucket fill shader.

luiq54:
I found a nice little hack that should hopefully fix #615 without needing to create separate shaders


OverloadedOrama (approved):
Looks good, thanks!

PR #630, opened 2022-01-06
Bugfix: Duplicate Shader in ShaderImageEffect before modifiying its code

mrtripie:
Fixes a bug mentioned by @Variable-ind in discord where modifying the code in the shader resource to add render_mode blend_premul_alpha gets accumulated on each use as it is modifying the original resource. (It also messes up the preview which shares the original shader resource.)


OverloadedOrama (approved):
Thanks!

PR #647, opened 2022-02-07
Move Guides without moving mouse outside canvas

Variable-ind:
Pan tool (**and Pan Tool alone**) is granted access to move guides without moving mouse outside canvas


OverloadedOrama (approved):
Thanks!

PR #648, opened 2022-02-10
improvements to copy\delete methods and (some) tag workflow

Variable-ind:
~~(Its still a work in progress till i comment on discord)~~
**To allow cloning/deleting an array of frames instead of just 1**

1. if a frame has at least one of it's cel selected then it will be cloned/copied
2. the cloned cels are intentionally placed immediately after the last selected frame
3. if the last selected frame is part of any tag then they would be included in that tag

**TAGS:**
1. fixed a bug when a tag of length 1 on frame 1 was removed in the presence of other tags
2. add and clone frames push the upcoming tags ahead and increase the size of tag they're in
3. ~~movement left or right now actively interact with tags~~ (may be made on another PR)

https://user-images.githubusercontent.com/77773850/153587819-71d5ea87-2ccf-4f3e-b94b-db39e5153b1b.mp4


Erevoid:
Hello there! From the changes you proposed we're going to implement the ability to clone/delete multiple frames and the displacement of the tags when a new cel is added before them. We already have something in mind for the rest of the suggestions you proposed in this PR and we believe they will be more user friendly. I recommend removing thοse from this PR for now until we discuss them ever further for future implementations.

PS: Keep in mind that this kind of PR which proposes lots of changes at once is not optimal because there's always a risk that a lot of work of yours will be lost. If you want to avoid that in the future try to keep your PRs even smaller, with only one change per PR, for everyone's convenience!

OverloadedOrama (approved):
Looks good, thank you!

PR #613, opened 2021-12-07
Replaces bucket same color pixels with a shader

novhack:
On my desktop reduces the process time for a 2048x2048 image from 16 seconds to 2.
I also plan to reuse this shader for the "indexed" mode of the palette system.

PR #649, opened 2022-02-14
Novhack fill shader

Variable-ind:
This Pull is the same as @novhack's Draft #613 but i think i fixed the bug (When pasting a pattern there are some unwanted pixels placed around the original image)

1. Also added a similarity feature that checks how much two pixels are similar/close together.


OverloadedOrama (approved):
Looks good to me. Thank you and novhack for the original PR! :)

PR #653, opened 2022-02-26
A simple id for extensions

Variable-ind:
A simple id system for extensions using the benefit of groups. it would suffice till a more advanced system is planned/implemented


OverloadedOrama (approved):
Thanks!

PR #654, opened 2022-02-26
Tag interaction

Variable-ind:
Tags now interact with movement buttons

https://user-images.githubusercontent.com/77773850/153587879-1a700c21-9ce4-42b9-bf35-d5e98099d198.mp4


Variable-ind:
But I'll still keep the code in my repo just in case 😃

PR #657, opened 2022-03-01
Draw optimizer

Variable-ind:
Should improve drawing on large size canvas significantly faster, similar to pull #554 and may solve issue #501. I would like it if @sayaks also get credited cause he was also working on the same issue 😁

### DIFF
--- a/src/Tools/BaseTool.gd
+++ b/src/Tools/BaseTool.gd
@@ -9,7 +9,7 @@ var cursor_text := ""
 var _cursor := Vector2.INF
 
 var _draw_cache: PoolVector2Array = []  # for storing already drawn pixels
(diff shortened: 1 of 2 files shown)


Variable-ind:
I decided to put a limit to the amount of cache stored. i ran some tests and think a value of 2000 will be decent enough.
## Demonstration
### Before
https://user-images.githubusercontent.com/77773850/156218194-c71df36a-28d2-4804-93de-e5c2fd0fa558.mp4
### After
https://user-images.githubusercontent.com/77773850/156218372-70a63c82-a705-4c97-a981-4448bfa7b36d.mp4

OverloadedOrama (approved):
Looks good to me, thanks!

PR #661, opened 2022-03-15
Optimize canvas texture updates

ArthyChaux:
In this line, a new texture is generated each time the canvas is updated it seems :
But it is mentionned in the documentation to use `void set_data(image: Image)` instead for slightly faster results, which updates the texture instead of creating a new one.

I hope that it is relevant !

Also, if it is approved, this change should be applied to everywhere it is used (except when initializing the texture)


OverloadedOrama (approved):
Looks good, thank you!

Issue #639, opened 2022-01-22
Flood fill issue


Distortions81:
**Pixelorama version:**
0.9.1 stable

**OS/device including version:**
Ubuntu 21

**Issue description:**
Bucket/Flood fill is insanely slow

**Steps to reproduce:**
Use flood fill on a blank 4096x4096 image.

MatteoPiovanelli-Laser:
I have attempted to implement the floodfill routine from the Allegro library (https://www1.udel.edu/CIS/software/dist/allegro-4.2.1/src/flood.c).
It seems to be somewhat faster than the current implementation in Bucket.gd it would replace.

I can provide a PR, but I'm not 100% confident I tested everything. Let me know if you are interested anyway, and if anyone can guide me making sure it's properly tested.

Some approximate common timings for flood fill of an entire picture on my machine for the different implementations follow. The columns labeled `(no color)` correspond to code where I commented out the calls to `_set_pixel` in order to only compare the algorithms.

| WxH | Bucket.gd (current) | Bucket.gd (no color) | Allegro (new) | Allegro (no color) |
| - | - | - | - | - |
| 512x512 px | 1570 ms | 1075ms | 910 ms | 330 ms |
| 1024x1024 px | 6250 ms | 4290ms | 3590 ms | 1300 ms |
| 2048x2048 px | 25000 ms | 17200 ms | 14400 ms | 5200 ms |

It's probably not a game changer. It's likely that looking at my code someone more experienced with gdscript might find better ways to optimize it.

OverloadedOrama:
@MatteoPiovanelli-Laser That sounds very interesting! Feel free to open a PR and we can review the code and test if it's working properly.

MatteoPiovanelli-Laser:
Further refactoring. The linter failed on a line that the parser had me change. I hope it's good for both now.

MatteoPiovanelli-Laser:
In case of specific image formats, it's possible to replace repeated calls to `image.set_pixel` by writing directly to the image's underlying buffer. That should speed things up further.

For example, see the `set data` column in the following table. The test are done for images in the `RGBA8` format, which it's my understanding is the default in Pixelorama, but I'm not sure whether it's the only one that' used.

| WxH | Bucket.gd (current) | Bucket.gd (no color) | Allegro (new) | Allegro (set data) | Allegro (no color) |
| - | - | - | - | - |
[...]

PR #667, opened 2022-03-30
Upgrade/floodfill

MatteoPiovanelli-Laser:
As discussed in #639 , this is my attempt at implementing the floodfill routine in gdscript

### DIFF
--- a/src/Tools/Bucket.gd
+++ b/src/Tools/Bucket.gd
@@ -150,13 +150,10 @@ func draw_start(position: Vector2) -> void:
 	):
 		return
 	var undo_data = _get_undo_data()
-	var start_time = OS.get_ticks_msec ()
 	if _fill_area == 0:
 		fill_in_area(position)
 	else:
 		fill_in_color(position)
-	var end_time = OS.get_ticks_msec ()
-	print("elapsed ",  end_time- start_time)
 	commit_undo("Draw", undo_data)
 
 
--- a/src/Tools/Bucket.gd
+++ b/src/Tools/Bucket.gd
@@ -9,6 +9,10 @@ var _fill_area := 0
 var _fill_with := 0
 var _offset_x := 0
 var _offset_y := 0
+# working array used as buffer for segments while flooding
+var _allegro_flood_segments : Array
+# results array per image while flooding
+var _allegro_image_segments : Array
 
 
 func _ready() -> void:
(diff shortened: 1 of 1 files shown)


MatteoPiovanelli-Laser:
Of course last night I forgot to check formatting guidelines and such. I'll take care of that after work. Sorry for that.

MatteoPiovanelli-Laser:
@OverloadedOrama sorry to ping you. I think with my updates this PR should be able to pass the checks in the automated workflows now, but it needs authorization for those to run.

OverloadedOrama:
> @OverloadedOrama sorry to ping you. I think with my updates this PR should be able to pass the checks in the automated workflows now, but it needs authorization for those to run.

My bad, I thought I had to authorize it once. Done!

MatteoPiovanelli-Laser:
I did more reformatting. Apparently I had been too zealous with going to new lines for function parameters, according to the parser.

MatteoPiovanelli-Laser:
Further refactoring. The linter failed on a line that the parser had me change. I hope it's good for both now.

MatteoPiovanelli-Laser:
I wish I could run the format/lint tasks on my machine, but there is an issue in gdtoolkit and their scripts error out on my pc.

OverloadedOrama:
> Further refactoring. The linter failed on a line that the parser had me change. I hope it's good for both now.

I think the reason it fails is because of this bug https://github.com/Scony/godot-gdscript-toolkit/issues/148. Best thing we can do for now is to add `# gdlint: ignore=max-line-length` in the previous line.



> I wish I could run the format/lint tasks on my machine, but there is an issue in gdtoolkit and their scripts error out on my pc.

Yeah unfortunately if I remember correctly it's not working properly on Windows.

MatteoPiovanelli-Laser:
The code finally passes all automated checks. Let me know any further improvements you require.

OverloadedOrama:
For the most part it seems to be working well, but there's weird behavior when using flood fill with non-rectangular selections.
This is what happens when I use the bucket tool on an elliptical selection.
image

While this is the expected behavior.
image

For some reason, it seems to only be filling the areas with the same width. I have observed this behavior only with selections, if we draw an ellipse and use the bucket tool inside it, the result is what we expect.

MatteoPiovanelli-Laser:
I'm checking now. The issue with the code in my PR is indeed on those lines. It's not correctly checking whether the segment above (or below) should be checked in cases where there is a selection area. Basically, it's testing above (or below) the left-most pixel of the segment.
I'm working on a fix.

MatteoPiovanelli-Laser:
The quick fix would be

```
segment.todo_above = position.y > 0
segment.todo_below = position.y < project.size.y - 1
```

That seems to cover it all. I'm working on something more precise.

MatteoPiovanelli-Laser:
I ended up going for that quick fix, with some comments explaining it, because it's just quicker than the alternatives I found.

OverloadedOrama (approved):
I've tested it and found no other issues. The performance increase is definitely there, so this seems to be good to merge. Thank you!

PR #669, opened 2022-04-05
Export conflict

Variable-ind:
Will give more info about exactly which of the two (path or name) are wrong


OverloadedOrama (approved):
Looks good, thanks!

PR #672, opened 2022-04-19
Further speed up of bucket fill

MatteoPiovanelli-Laser:
By cutting down on some tests that were being repeated after a point had already been found to be one to be colored, I made the flood fill a bit faster again.

### DIFF
--- a/src/Tools/Bucket.gd
+++ b/src/Tools/Bucket.gd
@@ -154,10 +154,13 @@ func draw_start(position: Vector2) -> void:
 	):
 		return
 	var undo_data = _get_undo_data()
+	var start_time = OS.get_ticks_msec ()
 	if _fill_area == 0:
 		fill_in_area(position)
 	else:
 		fill_in_color(position)
+	var end_time = OS.get_ticks_msec ()
+	print("elapsed ",  end_time- start_time)
 	commit_undo("Draw", undo_data)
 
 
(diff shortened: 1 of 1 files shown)


OverloadedOrama (approved):
I tried it and it seems to work great! I was under the impression that the only way the bucket fill would speed up like this would be using GDNative, and I am glad to be proven wrong. Thank you!

Variable-ind:
@MatteoPiovanelli-Laser I found a bug, pixels enclosed as shown here don't get effected by flood fill even if you click on top of it

https://user-images.githubusercontent.com/77773850/166229362-07a35e21-e221-42f4-bc80-5340d9d87bd1.mp4

MatteoPiovanelli-Laser:
I'm on the wrong machine to test this right now. Does the same issue happen regardless of whether you click to the right or to the left of the affected pixel? Looks to me I might have done a +2 where I needed to do a +1 or something like that where I'm computing where segments start. I'll see if I can check this later tonight.

Variable-ind:
yes, also i narrowed it down to this pull here: https://github.com/Orama-Interactive/Pixelorama/pull/681

Variable-ind:
i have narrowed it down further to line 305 of bucket.gd
is it safe to do `if segment.right_position >= segment.left_position:`
instead of `if segment.right_position > segment.left_position:`
cause that removes the bug

MatteoPiovanelli-Laser:
Yes, that actually makes a lot of sense. The code from my PR causes only segments at least two pixels long to be colored

MatteoPiovanelli-Laser:
@Variable-ind do you want to go ahead and do the PR with the fix or should I?

Variable-ind:
i think you should do it :smiley:
it's just a single line change right?
`if segment.right_position = segment.left_position:` to `if segment.right_position >= segment.left_position:`
in that case we could ask @OverloadedOrama to commit it internally (in that case the pull wont be necessary)

OverloadedOrama:
Fixed in 9e7dd1210b27aa83da0f955eaa73f0d61f44f17d

PR #673, opened 2022-04-20
Added shortcuts

Variable-ind:
Added *almost* each and every existing Shortcut in the Preferences Dialog (Also i removed one of the two fullscreen shortcuts in favor of this)


src/Autoload/Tools.gd
```diff
@@ -95,7 +95,7 @@ var tools := {
 		"""Hold %s to snap the angle of the line
 Hold %s to center the shape on the click origin
 Hold %s to displace the shape's origin""",
-		["Shift", "Ctrl", "Alt"]
+		["Shift", "configurable_ctrl", "configurable_alt"]
```
OverloadedOrama, line 98: Shouldn't this have "configurable_shift" instead of "Shift"?

Variable-ind: No because that shift's not influenced by input map and isn't configurable

Variable-ind: the "shift" represented here uses "event.shift" from "InputEventWithModifiers" (instead of the "shift" from "input map")

OverloadedOrama (approved):
Thank you!

PR #676, opened 2022-04-24
Rotation improvements

### LINKED ISSUE #675: Rotating image by multiples of 90 degrees is broken.
**Pixelorama version:**

0.10

**Issue description:**

Images rotated by 90, 180 and 270 degrees, are not correct pixelwise.

**Steps to reproduce:**
The issue is visible in the provided project. There are 2 triangles, both rotated clockwise by 90 degrees. Red, with pixelorama, and green, with Ubuntu's image viewer, with both results in separate layers. The triangles are made, so that the whole image should be covered, yet there are gaps in the red portion of the image.

I guess a simple override with a separate rotating function for those 3 rotations would fix it.

Also, I think the "Rotate Image" menu could be improved with 5 buttons for quicker use. These would be -90, -45, 0, +45, +90, with 0 setting the rotation degrees back to 0. This could be great for making top-down spritesheets quickly, and the addition of multiples of 45 degrees would be an improvement over the image viewer. But that's just my suggestion.

Variable-ind:
This pull addresses these two issues
1. The pivot was being set wrong in case of even lengths resulting in a 1px gap
2. The Image boundaries were being cropped during "scale_3x() method" in (upscale rotate downscale)

Old implementation rotated 90 degrees
Before

New implementation rotated 90 degrees
After

Maintainer edit: Fixes #675.


OverloadedOrama (approved):
Looks good to me, thanks!

PR #677, opened 2022-04-24
Shader-based gradients

### LINKED ISSUE #645: Gradient color picker insanely slow
**Pixelorama version:**

0.92

**Issue description:**

When using the gradient tool, and choosing 1 of the 2 colors, the color selector is very slow. Even more for larger images. The application remains frozen until it eventually catches up. This issue is also present in 0.91, but surprisingly NOT in 0.9, where it's very quick.

ALSO is the gradient preview, actually the same dimensions as the image? If yes, then maybe, for optimization, it could be replaced by a smaller image of the same proportions. (Like a 100x100 for any square image, since that's the max number of gradient steps.)

**Steps to reproduce:**
1. Create new image (Bigger dimensions like 512x512 make it more noticeable)
2. Select the gradient tool.
3. Click any of the 2 colors to open the color selector.
4. Move the RGB sliders, or wiggle the crosshair in the color square.

OverloadedOrama:
Gradient generation is now shader based. This results in much faster speeds compared to the previous GDScript generation algorithm.

Apart from the significant performance increase, the new gradient generation system offers 6 types of gradients. Linear, Radial, Step, Radial Step, Dithering & Radial Dithering. The previous gradient system only generated Step gradients. A lot more freedom is given to the users as well. Now, they can tweak the angle of the gradient, the size, position, radius and center for the radial gradients, and the dithering matrix for the dithering gradients.

https://user-images.githubusercontent.com/35376950/164997448-169e68ce-79f2-4b41-aae3-6705ee849082.mp4

Fixes #645, implements the first part of #628.


OverloadedOrama:
I managed to rewrite the Step & Dithering shaders (with their radial versions as well) without a for loop. This makes them work in the Web version with GLES2 and they should be a little more optimized now. I also rewrote the dithering textures loading logic, so that in the future we could allow the users to load their own dithering patterns, but that feature is outside the scope of this PR. I suppose some more UI/UX improvements are needed before merging.

OverloadedOrama:
That's a good point. I changed how the preview looks in all shaders except the two dithering ones, because they seemed to give different results than they do now. Although I am a bit worried that the other gradients may give different results as well, but so far I haven't noticed anything. Ideally, making the preview look the same as the output should happen outside of shader code, but I'm not sure how that would work.

AlphinAlbukhari:
Maybe using viewport with same size as canvas size?Place the preview inside there so we can see it being naturally pixelated

OverloadedOrama:
Should be good to merge now, if we want we can always make improvements later.

PR #679, opened 2022-04-27
Transparency improvements

Variable-ind:
This should remove the unwanted phenomena given below. i think it's caused due to division not giving precise numbers so an error margin is being taken into account

https://user-images.githubusercontent.com/77773850/165491423-165130d4-2bad-4c0c-ab2b-474a5c6dee29.mp4


OverloadedOrama:
I tested it but I can still see the issue when I maximize the window (1920x1080 monitor)
image

Variable-ind:
closing cause i was not able to solve this

PR #683, opened 2022-04-28
Shader based rotation

AlphinAlbukhari:
This will add a new shader based rotation mode, it's close to nearest neighbour but it uses GPU so it's much faster

https://user-images.githubusercontent.com/54466934/165759245-b01231e2-7af3-4735-96bd-c0f99ee52f5d.mp4


OverloadedOrama (approved):
Awesome, thank you!

PR #686, opened 2022-04-29
View Export dimensions

Variable-ind:
A simple feature that lets us view what the export size is in advance
Handy when exporting **Resized** images

https://user-images.githubusercontent.com/77773850/165977396-84784fae-4ccd-41dc-a1bd-a5da020fddea.mp4


OverloadedOrama (approved):
Thanks!

PR #688, opened 2022-05-01
[WIP] Layer Blending Mode

AlphinAlbukhari:
This breaking PR will rewrite Canvas rendering code by new shader-based rendering. This allows us to add layer blending modes, those are
- Normal
- Additive
- Subtract
- Multiply
- Screen
- Difference
- Burn
- Dodge
- Overlay
- Soft Light
- Hard Light

https://user-images.githubusercontent.com/54466934/166110404-c9218527-29ad-4fe8-84ca-c11478530152.mp4

**Changes**

- Rewritten layers rendering code, now layers are drawn by shader
  - This shader is generated by GDScript
  - It will generate only when you change the layer(s) such as adding new layer, removing, reordering, changing blend mode, and merging. Drawing, transforming, applying effects etc doesn't generate
  - Layers images and opacity are passed by uniform, so no need to generate everytime you change them
  - Those uniforms are updated everytime when Canvas _draw() is called
  - Additionally, you can generate shader with spesific layers by calling Canvas.generate_shader(layers: Array). This will return shader code (as a string of course). This func will only generate code, if you want to apply this to preview material just call Canvas.update_shader()
- No longer draw_texture(cel) on every preview, they only draw one for placeholder so we can apply the material
- Now previews has shared material, so when we change it on one preview, others will be changed too
- Layer merging now generates image with shader (using ShaderImageEffect), so blend mode is applied between them
  - This will generate shader with only two layers
- Exporter now also use shader as well, otherwise no blending

**Limitation**

Texture samplers (which are used to pass layer image) have limited amount, this depends on your hardware, some have 32, some have 16, 190, or maybe even more. The problem is when your layers count exceeds this limit, it will breaks.
We could use TextureArray but we'll have to use GLES3. What do you think?
**Update: Working on split-based approach, now we can have infinite layers. Still experimenting with it so it's not included in this branch yet.**

**Future**

This shader based rendering opens up so many possibilities, such as clip mask, filter mask, groups, and many more!

**Testing**

Since I have rewrote many things, this needs some testing.
And special thanks to @Variable-ind for testing and spotting many issues.

### DIFF
--- a/src/Autoload/Export.gd
+++ b/src/Autoload/Export.gd
@@ -386,7 +386,9 @@ func blend_layers(image: Image, frame: Frame, origin: Vector2 = Vector2(0, 0)) -
 	for i in Global.current_project.layers.size():
 		params["tex%s" % i] = frame.cels[i].image_texture
 		params["tex%s_opacity" % i] = frame.cels[i].opacity
-	gen.generate_image(cel_image, Global.canvas.material.shader, params, Global.current_project.size)
+	gen.generate_image(
+		cel_image, Global.canvas.material.shader, params, Global.current_project.size
+	)
 	image.blend_rect(cel_image, Rect2(Vector2.ZERO, Global.current_project.size), origin)
 	cel_image.unlock()
 	image.unlock()
--- a/src/Autoload/Export.gd
+++ b/src/Autoload/Export.gd
@@ -384,14 +384,13 @@ func blend_layers(image: Image, frame: Frame, origin: Vector2 = Vector2(0, 0)) -
 			var cel_image := Image.new()
 			cel_image.copy_from(cel.image)
 			cel_image.lock()
-			if cel.opacity < 1:  # If we have cel transparency
-				for xx in cel_image.get_size().x:
-					for yy in cel_image.get_size().y:
-						var pixel_color := cel_image.get_pixel(xx, yy)
-						var alpha: float = pixel_color.a * cel.opacity
-						cel_image.set_pixel(
-							xx, yy, Color(pixel_color.r, pixel_color.g, pixel_color.b, alpha)
-						)
+
+			var gen = ShaderImageEffect.new()
+			var params = {}
+			for i in Global.current_project.layers.size():
+				params["tex%s" % i] = frame.cels[i].image_texture
+				params["tex%s_opacity" % i] = frame.cels[i].opacity
+			gen.generate_image(cel_image, Global.canvas.material.shader, params, Global.current_project.size)
 			image.blend_rect(cel_image, Rect2(Vector2.ZERO, Global.current_project.size), origin)
 			cel_image.unlock()
 		layer_i += 1
--- a/src/Autoload/Export.gd
+++ b/src/Autoload/Export.gd
@@ -379,6 +379,9 @@ func get_proccessed_image_animation_tag_and_start_id(processed_image_id: int) ->
 func blend_layers(image: Image, frame: Frame, origin: Vector2 = Vector2(0, 0)) -> void:
 	image.lock()
 	var cel_image := Image.new()
+	cel_image.create(
+		Global.current_project.size.x, Global.current_project.size.y, false, Image.FORMAT_RGBA8
+	)
 	cel_image.lock()
 
 	var gen = ShaderImageEffect.new()
--- a/src/Autoload/Export.gd
+++ b/src/Autoload/Export.gd
@@ -378,22 +378,17 @@ func get_proccessed_image_animation_tag_and_start_id(processed_image_id: int) ->
 # Blends canvas layers into passed image starting from the origin position
 func blend_layers(image: Image, frame: Frame, origin: Vector2 = Vector2(0, 0)) -> void:
 	image.lock()
-	var layer_i := 0
-	for cel in frame.cels:
-		if Global.current_project.layers[layer_i].visible:
-			var cel_image := Image.new()
-			cel_image.copy_from(cel.image)
-			cel_image.lock()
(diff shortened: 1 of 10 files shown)


mrtripie:
I think we need to figure out both how to get around the limitations, and then how to combine this pr with my #689 Layer groups. As all the changes in this are in the same places Layer Groups changes, it will probably be confusing to try to rebase Layer Groups over this branch, so I think it may be easier to figure out which changes work best from this and manually insert them into the Layer Groups PR.

A few thoughts on the two:
- I need to make sure your blending equations properly work with the groups, I know that I took many attempts to figure out how to get the normal blend mode to work both without groups and with them
- The image effects filter previews work better than I thought they would  without applying it to each layer individually when doing multiple layers at a time, its a little inaccurate sometimes, but not bad. perhaps an "accurate" but slow mode could be added later. 
- I'm not sure if Canvas.gd is the best place for the shader generation to be, I added it in a separate file in my version (currently called LayerBlendShaderFactory.gd). I think that maybe that's better as it also needs to be used for generating shaders for group cel previews in the Animation Timeline. It also may have much more code put in it later for mixing image effects in it for FX Layers #619 or the "accurate" Image Effects filter preview I just mentioned
- I tried using a ViewportContainer over the viewport in Canvas.tscn in LayerGroups as an experimental optimization of rendering, but I believe it broke several things, and I'm unsure about how exactly it affected performance so I may remove that for now.

AlphinAlbukhari:
@mrtripie 

> The image effects filter previews work better than I thought they would without applying it to each layer individually when doing multiple layers at a time, its a little inaccurate sometimes, but not bad.

Can you show me when it's inaccurate?

> I'm not sure if Canvas.gd is the best place for the shader generation to be

Yeah i'd use separate Class or new singleton but that's too simple (at least for now)

mrtripie:
Different blend modes when using full frame HSV effect can sometimes result in off blending (because its only working on the final image for the preview), or multiple layers selected with selected cels option. Another example is having a semi transparent layer over another layer and using outline.

I wouldn't worry about it for now, as I think it won't be too bad most of the time, and will be difficult to fix (with worse performance) as you would have to combine the shader for the image effects and the layer blending shader. Probably best to add in a later pr, I'll probably work on it when I get to FX Layers #619  system, as i will need a system to combine the shaders anyway.

AlphinAlbukhari:
@mrtripie 
They worked fine for me?

https://user-images.githubusercontent.com/54466934/166206052-8032e9bc-72c6-4d13-b117-a5e5a5869289.mp4

mrtripie:
https://user-images.githubusercontent.com/65431647/166267051-f9f22338-585d-4cc1-aaaf-2a601b37576d.mp4

Currently the the preview is made by blending all used layers together (apparently reusing part of the export file functionality) and then applying the effect, instead of applying the effect to each layer and then blending together (which is how it would work when you apply the effect). Sometimes this is a little inaccurate.

For example: Adjust HSV with blend modes. Hue and Saturation seem correct (although maybe its just slightly off, or maybe it depends on the blend mode, I only tested a few).
Value shifts however can be noticeably off.

The other example is the semi transparent layer over other layers with the Outline effect, its only applied around the edges of everything in the preview, but you get overlapping outlines when applying to your project as each layer has its own outlines.

---
Again, this doesn't seem as bad as I thought it would be (the outline example seems like it wouldn't be used often, and is predictable, and the HSV only seems to be on value, and its probably not noticeable on small tweaks). Also certain effects like Mirror will never be affected by it.

Fixing it also seems like it would be quite challenging as you'd have to combine the effect shader and layer blending shader, or actually apply the affect to a copy of each layer and blend those (and applying expensive effects on many layers would be very slow, not smooth to adjust).

I think we should first get both blend modes and groups merged into Pixelorama, and then discuss for a later PR creating an "Accurate" preview mode for Image Effects that would be slower but apply the effect to each layer before blending. Perhaps including a "Hybrid" preview mode where it goes in "Fast" mode when adjusting sliders and "Accurate" when you're done could be nice too to get both performance and accuracy. My FX Layers #619  proposal will require the ability to mix the effects into the layers, so I might want to work on both of those things at the same time.

AlphinAlbukhari:
True, and the master branch would worked the same since it also uses final image. Maybe we could

1.  Use each layer's image, apply shader and generate 'shadered' image from every layer, then call Canvas.generate_shader() (now it will use custom textures, perhaps we will use your shader factory) to blend them into one shader. But since we're generating images from every layer, it will be slow
2. It's possible (and relatively easy) to combine shaders, without generating every shadered layer. We could generate another canvas shader, except with effects on every layer. I have the technique from my shader creation tool, the approach is by generating function on every layer, that function will have effects on it. For example
```
vec4 layer1(uv, {additional inputs here});
col.rgb = mix(col.rgb, {blending formula, with layer1.rgb}, layer1.a);
```

On layer1():
```
vec4 layer1(vec2 base_uv) {
 vec4 layer = texture(tex1, base_uv);
 layer.rgb *= 0.5; // let's say we're shifting value here
 return layer;
}
 ```


But that's right, before that we need to work on groups first

AlphinAlbukhari:
But adding those might will delay this PR, because it would be too much for RC release. So it's going to be on the next minor or major release

Variable-ind:
I also have a small suggestion. i need some way to view which layers have blending mode applied other than `normal` without having to check each layer cause it gets confusing...

Variable-ind:
Hy 👋, is it still in progress 😅? i noticed that last commit was like an year ago...

OverloadedOrama:
Superseded by #911. Thanks for the initial work and inspiration!

PR #690, opened 2022-05-02
Fix drop shadow

AlphinAlbukhari:
Currently drop shadow has issues

1. Incorrect blending because rgb blends with shadow color (as background, which is not a problem on opaque pixels)
2. Pixels on the edge are stretched

This PR will fix them.

Before:
image

After:
image

### DIFF
--- a/src/Shaders/DropShadow.tres
+++ b/src/Shaders/DropShadow.tres
@@ -2,6 +2,7 @@
 
 [resource]
 code = "shader_type canvas_item;
+render_mode unshaded;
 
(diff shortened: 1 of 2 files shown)


mrtripie:
Nice! I also noticed that currently only "Selected Cels" option is working properly, all the other options are working the same as "Selected Cels".

AlphinAlbukhari:
@mrtripie that's right, signal wasn't connected. Thank you for reporting 👍
image

OverloadedOrama (approved):
Looks good, thanks!

PR #693, opened 2022-05-06
Implemented text clipboard system instead of the old one

Variable-ind:
I changed the pull name from
`Implemented OS.clipboard (using var2str())` to `Implemented text clipboard system instead of the old one`

Uses ~~the `OS.clipboard`~~ a `text file` instead of the custom `Clipboard` class, thus allowing it to transfer/share images between multiple **"Pixelorama"** instances (Which is still _quite a lot of freedom and power_)
### Adresses:
https://github.com/Orama-Interactive/Pixelorama/discussions/402 the experience should now be the same as when pasting to a separate tab
> When opening the software and then opening another project (not from file>open) made with pixelorama, it opens two seperate windows instead of an extra tab. This makes it uncomfortable to copy and paste something between projects, if they are in a different window

### A Demonstration:

https://user-images.githubusercontent.com/77773850/167196576-1eb825fd-51ab-473f-94ca-d37c5d337c8d.mp4

### DIFF
--- a/src/UI/Canvas/Selection.gd
+++ b/src/UI/Canvas/Selection.gd
@@ -690,7 +690,11 @@ func copy() -> void:
 		"big_bounding_rectangle": cl_big_bounding_rectangle,
 		"selection_offset": cl_selection_offset,
 	}
-	OS.set_clipboard(var2str(transfer_clipboard))
+	# Store to ".clipboard.txt" file
+	var clipboard_file = File.new()
+	clipboard_file.open("user://clipboard.txt", File.WRITE)
+	clipboard_file.store_var(transfer_clipboard, true)
+	clipboard_file.close()
 
 	if !to_copy.is_empty():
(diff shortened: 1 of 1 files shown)


OverloadedOrama:
It looks good and it seems to be working well. I'm mostly worried that, since it is not a proper solution due to Godot's current limitations, it may confuse users if they see text that they don't remember copying in their clipboard content/history. Do you think it would be worth it to make this behavior toggleable, so users can choose between this and the old method?

TheFlowCorperation:
Mabey if there was a way when the programme is focused / refocused the os clipboard is cached and Godot str2var kicks in. The os clipboard then can be reset with the cached data.

Variable-ind:
II modified the code a bit. Now it doesn't use `OS.clipboard` nor the old `Clipboard` class but instead uses entirely new way... **Text Files*!!!

OverloadedOrama (approved):
I like this method. I tested it on HTML5 too and it seems to be working. Thank you!

PR #696, opened 2022-05-07
GIF export speedup

novhack:
I looked into LZW compression code in the gifexporter addon and simplified it a little bit.

The result is as speedup of around 3.6 times.

Using Roroto's awesome splashscreen it went from 43,26 seconds to 11,96 seconds in the editor. Compiled version of Pixelorama goes from 34.5 to 9.3 seconds.

I did some tests with multiple animation files and gifs exported from 0.10 and exported from this changed version are identical according to "cmp -l"

I will eventually commit this also to the jegor's repository but it seems to be outdated at this moment.

I also commented the decompress function because it now needs a rewrite and is currently unused.

There are for sure more ways the speed the whole thing up but small steps...

### DIFF
--- a/addons/gdgifexporter/gif-lzw/lzw.gd
+++ b/addons/gdgifexporter/gif-lzw/lzw.gd
@@ -3,51 +3,8 @@ extends Reference
 var lsbbitpacker = preload("./lsbbitpacker.gd")
 var lsbbitunpacker = preload("./lsbbitunpacker.gd")
 
-
-class CodeEntry:
-	var sequence: PoolByteArray
-	var raw_array: Array
-
-	func _init(_sequence):
-		raw_array = _sequence
-		sequence = _sequence
-
-	func add(other):
-		return CodeEntry.new(self.raw_array + other.raw_array)
-
-	func to_string():
-		var result: String = ""
-		for element in self.sequence:
-			result += str(element) + ", "
-		return result.substr(0, result.length() - 2)
-
-
-class CodeTable:
-	var entries: Dictionary = {}
-	var counter: int = 0
-	var lookup: Dictionary = {}
-
-	func add(entry) -> int:
-		self.entries[self.counter] = entry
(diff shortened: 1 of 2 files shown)


novhack:
Do not merge yet. I found a bug with big images which require quantization. They are exported as solid color gifs and I need to investigate.

novhack:
I rewrote the add_image_data_block() function to have less looping. It shaves of only around 100ms from the export I test on but it hugely helps in a profiler where it was causing big spikes.

novhack:
I am done with this PR for now. The next step would be to somehow speed up LSBLZWBitPacker.write_bits() function which gets called a lot but it already uses a lot of bithacks so there is not much room to work with.

addons/gdgifexporter/gif-lzw/lzw.gd
```diff
@@ -146,66 +103,65 @@ func compress_lzw(image: PoolByteArray, colors: PoolByteArray) -> Array:
...
-			# set PREVCODE = CODE
-			prevcode = code
-
-		# Detect when we should increase current code size and increase it.
-		var new_code_size_candidate: int = get_bits_number_for(code_table.counter)
-		if new_code_size_candidate > current_code_size:
-			current_code_size = new_code_size_candidate
-
-	return index_stream
+#func decompress_lzw(
```
OverloadedOrama, line 106: Might be best if we removed this method entirely instead of commenting it out.

novhack: I wanted to leave it as a reminder that it should be reimplemented with a proper gif import support but it's also possible to remove it.

novhack:
The last commit helped to shave off a few hundred milliseconds. Now the export takes around 11.5 seconds.
The next step is to add bulk export - allow multiple frames to be processed at the same time in a thread pool.

OverloadedOrama (approved):
Thank you!

PR #700, opened 2022-05-10
Implement the Keychain Plugin

OverloadedOrama:
Implements the [Keychain Plugin](https://github.com/Orama-Interactive/Keychain), that aims to improve the shortcut system by letting the users re-map every input action bindings available in the application. Compared to the previous shortcut settings, this new system allows for empty input actions, multiple input events (shortcut bindings) for every action, of multiple input types, such as Keyboard, Mouse and Gamepad buttons. As a result, Pixelorama should be a lot more configurable and accessible, and also a lot more usable in devices such as the Steam Deck, although I do not own one to confirm this.

Screenshot from Pixelorama

Implements most parts of #588, except the addition of missing shortcuts.

### Need to be taken care of before merging:
- [x] The first part of https://github.com/Orama-Interactive/Keychain/issues/5 needs to be fixed.
- [x] Make the plugin text translatable. This does not include translatable action names.
- [x] Tool modifiers, such as the shortcuts used for 1:1 shape creation, centered origin, displacement etc that were made configurable in #673, are currently not. They should be made configurable with different actions that are currently mapped to the same events, local/independent from each other.
- [x] Camera movement with arrow keys is configurable.

This PR does not come with any new input actions for elements that are missing, such as certain buttons and menu items that currently have no shortcuts. It also does not make stuff like moving the selection with arrow keys, and selecting multiple cels in the timeline configurable. These features are outside the scope of this PR and will be implemented at a later date.


OverloadedOrama:
I made the tool modifier shortcuts configurable, so this PR is now on-par with the master branch feature-wise. I think the PR is ready, extra features can be added after it gets merged. Needs some more testing and I will most likely merge it soon.

OverloadedOrama:
I wasn't originally intending to add the ability for users to create their own shortcut profiles in this PR, but I thought it might be best to do this as soon as possible, because I changed the way shortcut profiles (previously called presets) work so that each profile has its own .tres Resource file and this would break the previous shortcut settings users may have set.

So now we have support for multiple custom shortcut profiles. 🥳 But I don't think I will add anything else to this PR now, it should be good to get merged.

Issue #703, opened 2022-05-24
Some code gets executed twice when changing brush size


MatteoPiovanelli-Laser:
**Pixelorama version:**
284bf4bc2b76a870d27ada408d46857141283ec0

**OS/device including version:**
Windows 10 - Should be irrelevant.

**Issue description:**
![image](https://user-images.githubusercontent.com/19283945/170125691-3c7708cb-8eca-4496-9f4a-a1b210e36a8e.png)
When setting the size of a brush by either filling in the value in the SpinBox (A) or moving the HSlider (B) it will anyway go through `_on_BrushSize_value_change(value: float)` https://github.com/Orama-Interactive/Pixelorama/blob/284bf4bc2b76a870d27ada408d46857141283ec0/src/Tools/Draw.gd#L58
This will in turn call `update_config()`, that will ultimately update the "other" UI element (either the HSlider or the SpinBox, anyway the one the user hasn't directly interacted with), causing `_on_BrushSize_value_change(value: float)` to be invoked again.
The net effect of this is that a bunch of stuff is computed twice for no reason.

**Steps to reproduce:**
As described above.

**Proposed fix:**
Currently, this is `_on_BrushSize_value_changed(value: float)` here https://github.com/Orama-Interactive/Pixelorama/blob/284bf4bc2b76a870d27ada408d46857141283ec0/src/Tools/Draw.gd#L58
```
func _on_BrushSize_value_changed(value: float) -> void:
	_brush_size = int(value)
	_cache_limit = (_brush_size * _brush_size) * 3  # This equation seems the best match
	update_config()
	save_config()
```
It would be simple to prevent the double invocation of the functions down the line by checking if the value is actually changing
```
func _on_BrushSize_value_changed(value: float) -> void:
	if int(value) != _brush_size:
		_brush_size = int(value)
		_cache_limit = (_brush_size * _brush_size) * 3  # This equation seems the best match
		update_config()
		save_config()
```
I don't see particular drawbacks with this, but I'm not familiar with enough of the codebase to be sure.
Also, I don't know if similar "issues" may be found elsewhere with other UI elements that double down on a single value.

PR #704, opened 2022-05-29
Only update brush size config if size changed

MatteoPiovanelli-Laser:
This is the fix I proposed in #703 after noticing that the code in `update_config()` was being executed twice when changing the size of brushes.


OverloadedOrama (approved):
Looks good to me, thank you!

PR #705, opened 2022-05-29
Reworked `draw_tool`

MatteoPiovanelli-Laser:
This is has to do with #501

In draw.gd, refactored the `draw_tool(position)` function to memorize a few preliminary steps that may be skipped on every pixel.

This refactor is the easy part. I'll look into the rest, which may require breaking `_set_pixel` into a few different versions to handle the different cases and avoid repeating checks for every point.


MatteoPiovanelli-Laser:
For starters, I'll add a version of `_set_pixel` that doesn't use the `_draw_cache` buffer. From my tests, for log brush strokes using circle brushes, the code attempting to keep that updated costs almost 40% of the time used by `_set_pixel`.

MatteoPiovanelli-Laser:
In my last change, I made `draw_tool_circle` a bit faster by memorizing the displacements from the center of all points it should cause to color, but this is still visibly slow. The issue is that for large brushes it tries to set hundreds of thousands of pixels at a time, freezing the application temporarily.

I tried playing with dynamic step sizes in `draw_fill_gap`, but that was leading me down a rabbit hole of graphical artifacts and fine tuning, with no real performance improvement.

MatteoPiovanelli-Laser:
If I find the time/energy I'll look at how other graphics software implements the same kind of thing to see what's wrong here.

I also have a plan to replace the call to `_draw_tool` inside `draw_fill_gap` with something that only computes (and returns) the points that should be colored. Those would be added to an array, filtered to remove duplicates, and then the array would be iterated over to call `_set_pixel_no_cache`. My suspicion is that the "mistake" is in eagerly setting the pixels, rather than storing all coordinates and coloring everything at once. This may be similar to what we found when speeding up bucket fill.

I won't be able to do it tonight, unless I skip sleeping, but I may have time to play with it tomorrow.

MatteoPiovanelli-Laser:
Here's a screen grab from today's master:

https://user-images.githubusercontent.com/19283945/171647003-30425311-85bc-499b-b257-e75944dd02d4.mp4

And here's a screen grab using the code from this PR:


https://user-images.githubusercontent.com/19283945/171647080-b61ec3e6-65be-4df3-bd62-18265e2caacb.mp4

It's still not what I'd call good, but it's an improvement, I think.

OverloadedOrama (approved):
Looks good, thank you! Also yes, both the pencil and eraser tools (and the shading tool) extend Draw.gd, so their drawing code is pretty much the same.

PR #707, opened 2022-06-12
Add options for custom tile mode offest

GrantMoyer:
Adds options for custom tile mode using custom x and y basis for tile offsets. Custom offsets allow non-rectangular tiling, for example a hexagonal grid. By default, normal rectangular tiling is preserved. Drawing tools are also updated to correctly support tiling. Since tiles can now overlap, drawing tools draw on the tile with the nearest center.

I was not able to figure out how to get the bucket tool to work with the change. Currently it fills outside the nearest tile. I unsuccessfully tried modifying the flood fill algorithm, so any pointers to get it working would be appreciated.

Example of hexagonal tiling:

https://user-images.githubusercontent.com/6333127/173254834-174115d9-7d7a-4bd6-9ad1-5caa9d883283.mp4


GrantMoyer:
While I was re-organizing the change to please gdlint, I realized the tile mode offsets should probably be project specific instead of global settings. I'll tackle that over the next few days.

GrantMoyer:
Tile mode offsets are no project specific, and get saved into and loaded from project files.


https://user-images.githubusercontent.com/6333127/174421948-f2c07f1f-9ac5-4944-8463-788c849de97d.mp4

OverloadedOrama (commented):
It all looks good to me I think, great job! However, I think that the "Tile Mode Offsets" menu option should probably be moved to the View menu instead of the Image menu. The Image menu is for actions that change the image data, such as scaling, effects and filters, while the View menu is intended for stuff that change some visual aspects of the canvas, such as tile mode, grid, etc. I think it would also probably be best if TileModeOffsets.tscn and gd are moved away from ImageEffects, since they are not image effects (they don't affect the image data), nor TileModeOffsets.gd extends the ImageEffect class.

It would also be great to have a "Reset" button or something similar, so users can easily return to the original settings of tile mode.

GrantMoyer:
Let me know if you'd prefer the Tile Mode Offsets entry in the Tile Mode sub-menu.

OverloadedOrama (approved):
I can't find any other issues with this PR so I'd say it's good to get merged. The bucket tool issue is not a regression, it didn't work with tile mode before this PR either, so I think it's something that we can tackle in a future PR/commit. Thank you very much!

Variable-ind:
There is a bug the cursor draws even when it's outside the tiles (as if invisible tiles are there as well which is undesirable)

https://user-images.githubusercontent.com/77773850/180617490-0c678a28-6ae7-417f-98d5-576533001003.mp4

Variable-ind:
Okay i made a [pull](https://github.com/Orama-Interactive/Pixelorama/pull/723)  that addresses the issues

PR #711, opened 2022-07-01
Onion icon change

Variable-ind:
Changed the onion skin icon
the old one looked exactly like the play button so i changed it into something more intuitive
Screenshot from 2022-07-01 16-20-47


OverloadedOrama (approved):
Thanks!

PR #712, opened 2022-07-02
offset corrections

Variable-ind:
The snapping becomes weird if the "grid offset" is larger than the "grid size"
I fixed the bug by correcting lines 144 and 145 of SelectionTool.gd

(Note: i was only able to fix this for the selection code, the move tool still has this bug)


OverloadedOrama (approved):
Thank you!

PR #713, opened 2022-07-03
Fixed move offsets

Variable-ind:
Okay, so i modified it a bit,
1) I separated the offsetting of selection snapping from the move snapping (selection offsetting is done in `_input()` while move offsetting is done in `_snap_position()`)
2) Selection (in move tool) follows proper snapping while The normal move snapping are just increments of grid_size (Cause there is no proper definition for move snapping, i mean for a selection there is a clear corner of selection that can snap to grid even when multiple frames are selected, BUT for simple move snapping the concept of corner becomes "blurry" especially when one considers multiple selected frames, so i chose simple increments for move snapping cause it seems more logical "
3) And the user is given ability to align snapping with grid-lines themselves 😉")

**Old Implementation:**

https://user-images.githubusercontent.com/77773850/177038603-2bd28d0f-8661-4f5a-9465-ee0aaaa429a4.mp4

**New Implementation:**

https://user-images.githubusercontent.com/77773850/177038617-55490014-0f2b-4de8-8c5d-6506ab49b78b.mp4

### DIFF
--- a/src/Tools/Move.gd
+++ b/src/Tools/Move.gd
@@ -107,7 +107,7 @@ func _snap_position(position: Vector2) -> Vector2:
 		# The part below only corrects the offset for situations when there is no selection
 		# Offsets when there is selection is controlled in _input() function
 		if !Global.current_project.has_selection:
-			var move_offset :=  Vector2.ZERO
+			var move_offset := Vector2.ZERO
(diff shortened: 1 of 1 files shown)


OverloadedOrama (approved):
Thanks!

PR #723, opened 2022-07-24
Tile mode fixes (Heavy testing required)

Variable-ind:
The New Offset system for tiling also introduced a new series of bugs, (especially when two tiles overlapped)
This pull will address the following issues and will make the offset system **relatively** more stable. **Heavy testing is required for this pull**:
(All these changes are *spaghetti-linked* to one another and hence included in a single pull)
1. If mouse is above area overlapped by two tiles then the tile drawn above has higher priority
2. If mouse is above area overlapped by a central tile and another tile then the central/main tile is given priority
3. All the other tiles are drawn below the main tile
4. Fixed the issue of *"dark tiles"* (tiles appearing to be there even if no tile is there)
5. The offset dialog now shows only relevant information according to mode selected
6. Bucket tool still will only work on main tile (intentional)
7. Allow loading masks for custom tile (the default is already set to rectangle so users who want to use simple tiling won't need the set-up)

https://user-images.githubusercontent.com/77773850/180641196-eb270e9e-f281-491b-bcde-17f0da82e45f.mp4


GrantMoyer:
> 1. If mouse is above area overlapped by two tiles then the tile drawn above has higher priority
> 2. If mouse is above area overlapped by a central tile and another tile then the central/main tile is given priority
> 3. All the other tiles are drawn below the main tile

The existing behavior here is intentional. When the cursor is over multiple tiles, we want to draw to the tile whose center is nearest, because it works better in the case of non-square tiles. Compare the result of the two methods below (in both cases I drew on the top-left tile):

| Draw to nearest tile | Draw to topmost tile |
|:----:|:----:|
| border_nearest | border_topmost |

By drawing to the nearest tile, we ensure none of the pixels are drawn outside the non-square tile. Drawing to the topmost tile leads to unintuitive results, since the ordering of tiles is arbitrary.

Variable-ind:
@GrantMoyer Could you send a video so i can understand better?

Variable-ind:
@GrantMoyer i checked... it seems nearest tile method is working for hexagonal tiles but completely fails for isometric ones
Try this example with both nearest and top-most methods
[Test.pxo.zip](https://github.com/Orama-Interactive/Pixelorama/files/9175954/Test.pxo.zip)

Variable-ind:
@GrantMoyer This is the issue i wanted to fix (i could only think of the top tile method as a suitable workaround :sweat_smile:), if you another way this can be fixed without using it then feel free to create a pull for it :smiley: :heart:

GrantMoyer:
I see. Using euclidean distance chooses the wrong nearest tile for dimetric tiles, because they're really in a projected space.

I'm not sure if there's a good general solution, or if we'll just have an explicit option for hex/diamond tiles vs dimetric tiles. I'll have to think about it for a little bit.

Variable-ind:
Sure 😃, i'll keep it as a draft for now till the discussion is resolved 👍

GrantMoyer:
The solutions I've thought of are listed below. For completeness, I also list some non-solutions.

#### General solutions

- Project cursor and tile positions onto user defined 3D plane, find the nearest tile in that plane

  This would be too abstract for users without a math background to configure. I've only included it in the list for completeness.

- Use an explicit mask layer to determine if a pixel belongs to a tile (the mask would look like a blank version of the filled in tile)

  This is the most general solution, and it would allow whatever wacky tiling a user wants, including non-contiguous tiles. However, it would also involve more manual setup. It would also require significant changes to the code.

- Prefer drawing to pixels which are already non-transparent.

  This would require the users to first draw a blank version of the tile. It's hard to tell without trying it out, by my instinct is that this would be a pain to work with. It's also unclear what should happen if all of the overlapping tiles have a transparent pixel, or multiple overlapping tile have a non-transparent pixel, and it's unclear if the bucket tool could eventually be extended to work with this option.

#### Special case solutions

- Add an explicit options for square, dimetric, and hex tiles to the tilemode offsets options.

  This option would be fairly quick and simple for users to configure. It wouldn't support some unusual tilings, like trimetric projection, but that might not be a tiling anyone will ever need in the first place. Diamond tiles are also supported; they're a special case of dimetric tiles. The code changes would be similar to the existing changes in this PR.

#### Non-solutions

- Maintain the current behavior (draw to the nearest tile)

  The current behavior does not work for dimetric tiles

- Draw to the highest layer tile

  This often draws to the "wrong" tile when drawing on lower layer tiles. For instance, at 19 seconds in this PR's first video attachment, the pencil tool draws into the transparent space instead of onto the existing part of the tile.

I think the special case solution is probably the best option. It's the most straightforward for the end user to configure, and I think it's unlikely the extra flexibility afforded by the general solutions would ever be used by anyone. We may also wish to add a way to quickly toggle between the tile choosing behavior and the drawing to the highest layer tile behavior, for instance when a user wants to draw grass sticking out of the top edge of a dimetric tile.

Variable-ind:
@GrantMoyer After reading the suggestions i think this would be ideal:
1) We merge the top-tile mode for now as a bandage cause at the moment it's the most ideal of the two
2) Then you can decide and create a pr for a more proper solution (the above message will be moved to discussion for safe-keeping)

Variable-ind:
@GrantMoyer I tried out the masking system, the mask is a rectangular tile by default but we can load isometric masks as a .png image (Yes in theory it should work for all tiles)
the video shows comparison with a rectangular mask and when an isometric mask is loaded (i've also included them here)
[Test Project.zip](https://github.com/Orama-Interactive/Pixelorama/files/9176419/Test.Project.zip)

https://user-images.githubusercontent.com/77773850/180661123-6e812c30-1725-4f86-8b56-ae143ec1cbf6.mp4

Variable-ind:
And here's the result with Hexagonal (Pretty decent right?)
[Test Hexagonal.zip](https://github.com/Orama-Interactive/Pixelorama/files/9176454/Test.Hexagonal.zip)

https://user-images.githubusercontent.com/77773850/180661471-530ea121-9a8d-46eb-9563-6eca268541f0.mp4

Variable-ind:
Okay, now that I've added the masking solution. I'll just go and mark it ready for review :smile:

src/Classes/Tiles.gd
```diff
@@ -41,29 +44,38 @@ func get_bounding_rect() -> Rect2:
...
+	var positions = Global.canvas.tile_mode.get_tile_positions()
+	positions.append(Vector2.ZERO)
+
+	var candidates := []
+	for pos_ind in positions.size():
+		# Tiles on top gets detected first in case of overlap
+		var pos = positions[positions.size() - pos_ind - 1]
+		var test_rect = Rect2(pos, tile_size)
+		if test_rect.has_point(point):
+			candidates.append(test_rect)
```
GrantMoyer, line 56: Since you sort candidates later, I don't think you need index from the end:

```suggestion
	for pos in positions:
		# Tiles on top gets detected first in case of overlap
		var test_rect = Rect2(pos, tile_size)
		if test_rect.has_point(point):
			candidates.append(test_rect)
```

src/Classes/Tiles.gd
```diff
@@ -41,29 +44,38 @@ func get_bounding_rect() -> Rect2:
...
+	if candidates.empty():
+		return Rect2(Vector2.ZERO, tile_size)
+
+	var final := []
+	for candidate in candidates:
+		var rel_pos = point - candidate.position
+		tile_mask.lock()
+		if tile_mask.get_pixelv(rel_pos).a == 1.0:
+			final.append(candidate)
+		tile_mask.unlock()
```
GrantMoyer, line 66: I think you can avoid some overhead by locking/unlocking tile_mask outside the loop. You should merge this with the previous loop too, and filter the positions in one pass.

src/Classes/Tiles.gd
```diff
@@ -76,15 +88,11 @@ func get_canon_position(position: Vector2) -> Vector2:
...
-		_:
-			return nearest_tile_tile_space == Vector2.ZERO
+	var positions = Global.canvas.tile_mode.get_tile_positions()
+	positions.append(Vector2.ZERO)  # The central tile is included manually
+
+	for pos in positions:
+		var test_rect = Rect2(pos, tile_size)
+		if test_rect.has_point(point):
+			return true
+	return false
```
GrantMoyer, line 98: I think we may also need to test against the mask here.

src/Classes/Tiles.gd
```diff
@@ -7,12 +7,15 @@ var mode: int = MODE.NONE
 var x_basis: Vector2
 var y_basis: Vector2
 var tile_size: Vector2
+var tile_mask := Image.new()
```
GrantMoyer, line 10: We should probably persist the mask to the project file if a specific mask is loaded.

GrantMoyer (commented):
I think the mask is a good stopgap solution to support both isometric and hexagonal tiles. I'll probably take a swing at a more user friendly interface at some point though.

src/Classes/Tiles.gd
Variable-ind: Done

src/Classes/Tiles.gd
Variable-ind: not sure what you mean here...

GrantMoyer: For example, see the serialize function in Project.gd (line 296). We store the tile offsets in the project file so that the offsets are remembered when the project is loaded later. We should probably do the same for the mask, because otherwise the user will need to load the mask every time they load the project.

Variable-ind: Done :+1:
I also modified the dialog (just a bit...)
Hope it's good enough
Screenshot from 2022-07-27 16-25-13

src/Classes/Project.gd
```diff
@@ -354,6 +354,8 @@ func serialize() -> Dictionary:
 		"name": name,
 		"size_x": size.x,
 		"size_y": size.y,
+		"has_mask": tiles.has_mask,
+		"tile_mask": var2str(tiles.tile_mask),
```
OverloadedOrama, line 358: The issue with this is that even if there is no mask, a full white image with the size of the project will get saved in the pxo, taking unneeded space. It would be better if this line gets removed and, just above `return project_data` this condition gets added:
```
if tiles.has_mask:
		project_data["tile_mask"] = var2str(tiles.tile_mask)
```

I'm also not sure if var2str is a good idea, this saves the image data as a string instead of binary, which may take more space than binary does, but I haven't tested it so I am unsure. The cels and project brushes themselves get saved as binary data, however.

OverloadedOrama (changes requested):

OverloadedOrama (approved):
I think this looks good for now, and I did not encounter any other issues. Perhaps it could be a bit user friendlier like GrantMoyer said, an idea I had was the ability to use one of the cels as the mask, if the user doesn't want to load an external image. But this is something that we can look at later in a discussion or a future PR.


Thank you!

PR #724, opened 2022-07-24
Pivot remembering and algorithm re-arrangement

Variable-ind:
The Pivot will now remember it's last position
Algorithms are also re-arranged according to their speed (shader is the new default)

https://user-images.githubusercontent.com/77773850/180655267-ddaa8a71-2895-47c7-9742-a762a96ce530.mp4


OverloadedOrama (approved):
Thank you!

PR #728, opened 2022-07-28
removed initial angle quick rotation buttons

Variable-ind:
This pull is an attempt to make Rotate dialog UI a bit cleaner by removing the quick rotation buttons for "initial angle" cause they may not have much use anyway

Screenshot_from_2022-07-28_12-49-08


OverloadedOrama (approved):
Thanks!

PR #729, opened 2022-07-29
Fix (slightly) wrong color being drawn when selected from palette

Variable-ind:
i noticed that some of the colors drawn when selected from palette was slightly wrong (about 1px difference in either R or G or B values)
this pull attempts to fix this issue

#### Fixing this should now pave the way for some cool stuff like indexed mode and other things


OverloadedOrama (approved):
Thank you! We should probably keep this issue in mind though in case we find a better solution, but this seems to be working for now.

PR #730, opened 2022-07-29
Highlight color in palette if it's selected in color slots

Variable-ind:
currently the palette color swatch is highlighted only when we click on it, and when we change color and select the old color again (using color picker or any other method) the palette swatch is not highlighted.
This pull fixes this
(This pull relies on pull #729 to be merged before it)

https://user-images.githubusercontent.com/77773850/181689507-48815c48-0d14-4e0b-8fab-dd5b8454c8d8.mp4


OverloadedOrama (approved):
Looks good, thanks!

PR #731, opened 2022-08-02
Change tilemask loading

Variable-ind:
Load tilemask from current frame instead of selecting from FileDialog


OverloadedOrama:
Hmm, I just realized that there is no way to clear the tile mask, if the user doesn't want to use a mask anymore. That's relevant for the previous PR too but I forgot about it. I think once this is added, this PR should be ready to get merged.

Variable-ind:
if the user attempts to make a tilemask out of an empty frame then the tilemask will reset :smile:

OverloadedOrama (approved):
Thanks!

PR #733, opened 2022-08-06
New theme-related methods to ExtensionsAPI

Variable-ind:
1. Added new `set_theme` and `find_theme_index` methods to ExtensionsAPI
2. Make **HandleThemes.gd** wait in case themes are being loaded from extensions
3. Hide the accidentally visible **TileModeOffsetsDialog.tscn**
4. update readme for Godot 3.5

### DIFF
--- a/src/Autoload/ExtensionsAPI.gd
+++ b/src/Autoload/ExtensionsAPI.gd
@@ -192,7 +192,7 @@ func add_theme(theme: Theme) -> void:
 	themes.add_theme(theme)
 
 
-func find_theme(theme :Theme) -> int:
+func find_theme(theme: Theme) -> int:
 	var themes: BoxContainer = Global.preferences_dialog.find_node("Themes")
 	return themes.themes.find(theme)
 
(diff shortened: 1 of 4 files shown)


src/Autoload/ExtensionsAPI.gd
```diff
@@ -192,9 +192,23 @@ func add_theme(theme: Theme) -> void:
...
+func find_theme(theme: Theme) -> int:
+	var themes: BoxContainer = Global.preferences_dialog.find_node("Themes")
+	return themes.themes.find(theme)
+
+
 func get_theme() -> Theme:
 	return Global.control.theme
 
 
+func set_theme(idx: int) -> int:
```
OverloadedOrama, line 204: Nitpicking I suppose, but since the only possible return values are two, OK and -1, maybe it could return a bool instead, that is true or false depending if it has failed or not.

I'm also thinking if `find_theme()` would be better if renamed to `find_theme_index()` to make it a bit more clear what it does, but I'm not sure if that's needed.

Variable-ind: I use OK cause this system is conventional with error detection of godot
Yeah I'll change the name in a moment

Variable-ind: Done :+1:

OverloadedOrama (approved):
Looks good, thank you!

PR #734, opened 2022-08-07
Make changing window transparency only affect the main canvas' transparency

Variable-ind:
Only make the Tabcontainer **With** the "Main Canvas" change its modulation


OverloadedOrama (approved):
Thanks!

PR #754, opened 2022-09-20
Remember last choice in import options

Variable-ind:
Implements 2nd point of  #735

> Import dialog should remember the last option. I was importing separate frames of animation and I had to always re-select New frame.


OverloadedOrama (approved):
Looks good, thanks!

PR #764, opened 2022-10-07
Linked Cel Refactor and Timeline Refactor Tweaks

mrtripie:
Implemented #719 
- New concept of a "link set":
  - Layer's linked_cels array of Frames was replaced with a cel_link_sets array of array of Cels
    - **Each inner array represents a "link set", allowing for a single layer to have multiple different sets of linked cels**
    - The Frames were changed to Cels which makes certain things easier and others harder, but I think makes more sense
  - Each Cel has a reference to its link_set array, if linked, or null if not
  - The linked cel indicator on Cel Buttons is now a different hue depending on the index of the Cel's link_set inside the Layer's cel_link_sets array.
    - An issue with this is that when a link sets are added/removed to cel_link_sets, other link_set's hue will change. (and when undoing and its removed/added back it will likely be a different color than before)
      - If it seems worth it, this could be solved by representing link sets with a simple class or Dictionary containing an Array and hue float, and setting automatically the hue of the link set when its created.
- Linked Cel data was moved from PixelLayer to BaseLayer to make it simpler for multiple layer types
- **Deleted Project.duplicate_layers()**
- **Enabled drag and drop to move/swap linked cels on the same layer**
- **Can now link/unlink multiple cels at a time**
- Enabling new_cels_linked no longer links the current frame's Cel
- When new_cels_linked is enabled, new Cels will be linked to previous frame's Cel (the Cel to the left of the new one)
- Updated the tooltip for PixelLayerButton's LinkButton to be more accurate (it doesn't enable/disable Cel linking, it enables/disables automatic linking of new Cels when creating a new frame)

# Timeline Refactor Tweaks:
- **Rewrote the copy methods on BaseLayer to use get/set_content, as they now should work on any layer type (without overriding) and are only used in one place, removed the methods and inlined where used)**
- Added a texture argument to set_content

# Other:
- Bug fixes from my timeline refactor
- General cleanup

### DIFF
--- a/Translations/Translations.pot
+++ b/Translations/Translations.pot
@@ -1540,10 +1540,10 @@ msgstr ""
 msgid "Add Frame Tag"
 msgstr ""
 
-msgid "Unlink Cels"
+msgid "Link Cels to"
 msgstr ""
 
-msgid "Unlink Selected Cels"
+msgid "Unlink Cels"
 msgstr ""
 
 msgid "Frame Properties"
--- a/Translations/Translations.pot
+++ b/Translations/Translations.pot
@@ -1540,10 +1540,10 @@ msgstr ""
 msgid "Add Frame Tag"
 msgstr ""
 
-msgid "Link Cel"
+msgid "Unlink Cels"
 msgstr ""
 
-msgid "Unlink Cel"
+msgid "Unlink Selected Cels"
 msgstr ""
 
 msgid "Frame Properties"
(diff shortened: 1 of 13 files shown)


Translations/Translations.pot
```diff
@@ -1540,10 +1540,10 @@ msgstr ""
 msgid "Add Frame Tag"
 msgstr ""
 
-msgid "Link Cel"
+msgid "Unlink Cels"
 msgstr ""
 
-msgid "Unlink Cel"
+msgid "Unlink Selected Cels"
```
OverloadedOrama, line 1546: Is this string used somewhere? Shouldn't this be "Link Cels to"?

mrtripie: Its inside the right click menu on cel buttons.

mrtripie: Actually, it seems like this got messed up somehow, I'll update it in a minute

mrtripie: Should be fixed now

OverloadedOrama (approved):
Looks good to me. I tested it too and couldn't find any issues, so it should be good to merge. Thank you!

PR #765, opened 2022-10-14
Fix id_map error on open_pxo_file as well as grid being wrong size

mrtripie:
This fixes this error that showed up in the #698 PR when opening a pxo file (without any changes to the empty default file):
```
E 0:00:11.149   getornull: Condition "!id_map.has(p_rid.get_data())" is true. Returned: nullptr
  <C++ Source>  ./core/rid.h:151 @ getornull()
```
This was caused by Global.canvas.update() not being called

Also fixes the grid not being updated, and maybe other bugs.

Not sure if just using the Project.change_project() will cause other issues though


OverloadedOrama (approved):
Thanks!

PR #766, opened 2022-10-16
Keep colors on cel link sets

mrtripie:
- Changed link sets from an Array to a Dictionary with { "cels": Array, "hue": float }
- Hue is calculated by finding the largest gap in hue from the layer's current link sets


OverloadedOrama (approved):
Looks good, thanks!

PR #772, opened 2022-10-27
APNG Exporter

20kdc:
This allows exporting animations in the APNG format (which doesn't have the palette or transparency limitations of GIF).
It also cleans up the internals to allow adding more animation export formats easily.

The main UI change is making the export file format selector actually usable to select file formats.

Test example of exported output:
loader

### DIFF
--- a/project.godot
+++ b/project.godot
@@ -15,6 +15,11 @@ _global_script_classes=[ {
 "path": "res://src/Classes/AnimationTag.gd"
 }, {
 "base": "Reference",
+"class": "BaseAnimationExporter",
+"language": "GDScript",
+"path": "res://src/Classes/AnimationExporters/BaseAnimationExporter.gd"
+}, {
+"base": "Reference",
 "class": "BaseCel",
 "language": "GDScript",
 "path": "res://src/Classes/BaseCel.gd"
@@ -48,6 +53,11 @@ _global_script_classes=[ {
 "class": "Frame",
 "language": "GDScript",
 "path": "res://src/Classes/Frame.gd"
+}, {
+"base": "BaseAnimationExporter",
+"class": "GIFAnimationExporter",
+"language": "GDScript",
+"path": "res://src/Classes/AnimationExporters/GIFAnimationExporter.gd"
 }, {
 "base": "Control",
 "class": "GradientEditNode",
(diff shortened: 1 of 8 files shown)


luiq54:
Hello! I request you to move this code into an Godot addon so that it can be maintained independently from Pixelorama and will also help other Godot Projects! You can reference the GIF exporter addon as template.

20kdc:
If I'm moving the APNG export code into a Godot addon, does that mean I should migrate it to https://gitlab.com/20kdc/20kdc_godot_addons ? How does upstreaming/downstreaming interact with Pixelorama?
Or do I simply need to separate it out for file organization purposes?

OverloadedOrama:
If you want to make it a separate addon so other Godot projects can use it, you can put it on a repository on your own wherever you want to host it. There is no automatic downstream interaction, we basically just copy the files from the repository into Pixelorama's repository inside the `addon` folder, pretty much the same way as to how addons are being downloaded from Godot's Asset Store and integrated in the project's files, and we put the licenses and version for each addon in we use in the [readme file inside the addons folder](https://github.com/Orama-Interactive/Pixelorama/blob/master/addons/README.md).

Of course, this is optional and we only recommend doing it so other projects can use the exporter. If you don't want to make it an addon of its own, the current file organization works just fine.

20kdc:
If the APNG exporter were more than a single file, I'd agree with the assessment to put it into an addon.

As it is, however, I feel like more effort would be spent providing the necessary separation between addon and Pixelorama than it would take for anyone interested to copy it out.
I made sure that it's not particularly coupled to Pixelorama in the first place.

If there is a method I would recommend to achieve this, it would be to work with `gdgifexporter` to work out how to make an abstraction layer between formats at that addon's level, and therefore expand that addon's scope to include APNG.

20kdc:
So hopefully I have managed to fix the format/linting issues with this PR

OverloadedOrama (approved):
Looks good to me. I tested it a bit and couldn't find any issues, so I think it should be good to merge. Don't worry too much about formatting/linting, if it's an issue I can handle that for you, although the errors seem to have been fixed now.

Thanks! I will review your other pull request soon as well.

PR #781, opened 2022-11-21
Export dialog UX overhaul

OverloadedOrama:
Implements https://github.com/Orama-Interactive/Pixelorama/discussions/605#discussioncomment-4190666. Deviates from the original proposal and does not implement all of the new suggested features yet.

image

This PR makes the export dialog more intuitive and simpler to use by merging the "Frame" and "Animation" tabs into one. Now, all frames will be exported as multiple png files by default, with the ability to choose only to export the currently selected frame(s), thus keeping the functionality of the previous "Frame" tab and only exporting one png file, or multiples if there are more than one selected frames. To export a .gif or a .apng file, simply change the file extension.

image

The PR also adds some new features:
- Specific layer exporting. Choose between all visible layers, selected layers, or specific layers. Also supports exporting group layers that blend their child layers. _Kind of_ implements #773.
- A new CollapsibleContainer custom node, which is being used to hide advanced settings. This node can be used anywhere in the UI, not just the export dialog.
- Specific frame exporting for multiple pngs, gifs and apngs, besides just spritesheets which is the case in the current master branch. Should fully implement #472.
- Animation direction now supports multiple pngs and spritesheets besides just gifs and apngs.

Future improvements that can be made but are outside the scope of this PR:
- Extra options for the spritesheets, such as a gap between frames.
- Add a "Overwrite all existing" on the overwrite dialog when exporting to multiple files. (#585)
- Add a way to choose multiple tags and layers instead of just one.
- Add an option to export multiple spritesheet files for each tag.
- Add an option to export each layer as a separate file. (#773)

The PR should be ready to get merged, but feedback on the UX as well as testing would be very welcome!

### DIFF
--- a/Translations/Translations.pot
+++ b/Translations/Translations.pot
@@ -441,13 +441,30 @@ msgstr ""
 msgid "Open a Directory"
 msgstr ""
 
-msgid "All frames as multiple files"
-msgstr ""
-
-msgid "All frames as a single file animation"
-msgstr ""
-
 msgid "Background:"
+msgstr ""
+
+#. Found in the export dialog
+msgid "Selected frames"
+msgstr ""
+
+msgid "Layers:"
+msgstr ""
+
+#. Found in the export dialog
+msgid "Visible layers"
+msgstr ""
+
+#. Found in the export dialog
+msgid "Selected layers"
+msgstr ""
+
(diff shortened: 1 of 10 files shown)


OverloadedOrama:
Updated the PR to add a new CollapsibleContainer custom node, that is being used to contain the advanced settings.

PR #784, opened 2022-11-30
Tool shortcuts should ignore macOS Command modifier.

### LINKED ISSUE #778: macOS hotkeys sometimes change tools.
**Pixelorama version:**
0.10.3-stable

macOS 13

**Issue description:**
Using macOS the hotkey "CMD+C" (copy) will cause the current tool to change to the ellipse tool.

**Steps to reproduce:**
Use the rectangle tool to select a region of the image, and press "CMD+C" to copy that region. The current tool will change to the ellipse tool.

radicaled:
This prevents common shortcuts like CMD+S (Save), CMD+C (Copy), etc, from activating a tool.

Fixes #778


kees-:
Thanks!

PR #787, opened 2022-12-04
Added a preview to layouts

Variable-ind:
Added a decorative preview to layouts, that gives you an idea of what the preview looks like

keep in mind it's only a preview and won't change the actual layout, i think the current method is more efficient (i-e create a new layout and delete the old layout)

https://user-images.githubusercontent.com/77773850/205478749-1cd98119-641f-4eba-872a-1cdeb366980b.mp4


OverloadedOrama (approved):
Looks good to me, thank you!

PR #788, opened 2022-12-05
Decorate headers

Variable-ind:
1) Added a header to the reference images pannel
2) and did some changes to the tools and reference image header

### Old:
image

### New:
image


OverloadedOrama (approved):
Thanks!

PR #789, opened 2022-12-06
Reference UI improvements

Variable-ind:
Improved the Reference image panel UI
This pull includes:

1. Better alignment of Position, Scale, Opacity Sliders
2. The text of "Remove" button is Red
3. Clicking on the "Path" now opens the folder in which reference image is placed (disabled for HTML)
4. A new Image Preview
5. A new option to enable or disable filter (Filter is off by default)
6. A warning appears if the image is not found

### Before
Screenshot from 2022-12-06 16-29-30

### After
Screenshot from 2022-12-06 16-31-03

### DIFF
--- a/src/UI/Canvas/ReferenceImage.gd
+++ b/src/UI/Canvas/ReferenceImage.gd
@@ -10,6 +10,7 @@ var image_path: String = ""
 
 var filter = false
 
+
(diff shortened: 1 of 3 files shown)


OverloadedOrama (approved):
Looks great, thank you!

PR #792, opened 2022-12-08
Paint selection brush

Variable-ind:
Implemented the **Paint Selection Tool** from #567.
It works the same way as the other selections (shift to add, ctrl to subtract, etc...)
But in this tool you can select things by drawing over them

### Things i won't do for this pull (but are required)
there are some things that were too complex for me which are
1. Integrating it with the Keychain plugin (i-e Creating the tool's Shortcut)
2. ~~The icon is in desperate need of a re-design~~ (I re-designed it a bit but further changes to icon can be made if required)

### Demonstration
https://user-images.githubusercontent.com/77773850/206540446-2fcc0d04-5b92-400a-8240-00d378f91a8a.mp4


OverloadedOrama (approved):
Looks good to me, thank you!

PR #797, opened 2022-12-22
APNG Loading

20kdc:
This adds an APNG loader. Loaded APNGs are treated as opened projects right now, as importing them "as an image" doesn't really make sense.

Due to the "APNGs are subsets of PNGs" rule, an APNG will only be considered as such if it contains at least one APNG frame (which makes it distinct from a regular unanimated PNG).

On the previous PR regarding APNG, https://github.com/Orama-Interactive/Pixelorama/pull/772 @luiq54 mentioned as so:

> Hello! I request you to move this code into an Godot addon so that it can be maintained independently from Pixelorama and will also help other Godot Projects! You can reference the GIF exporter addon as template.

With the addition of an APNG loader that shares quite a bit of common code with the saver (due to the nature of the technique used to load APNG efficiently with very little code), this has become rather important.

With the original APNG saver this was unnecessary as it was a single file in size, but that has now changed.

As such, I have refactored my APNG saver into an addon, and included an APNG loader.

The GIF exporter integration code still remains in Pixelorama (that part of the structure exists to keep the code clean more than anything else).

My addon is licensed under the Unlicense as it has been recommended to me as a better license for code than CC0 (due to Fedora policies and so forth).

I would recommend Pixelorama doesn't consider upstreaming to be much of a requirement for changes to the addon, my addon doesn't have much in the way of support guarantees, it only exists in this form because it's a good point that having it like this helps other Godot projects.

The following file loads in Pixelorama with this PR:
import_demo

Notably, this doesn't include a GIF importer, but it does provide a useful path to import a GIF in that ffmpeg is capable of converting a GIF to an APNG (previously there was no way to import a GIF at all that I'm aware of without somehow getting it into a spritesheet).

Also notably, the addon may be of use for future animated splash screens, as it is capable of importing an APNG as an AnimatedTexture.


src/Classes/AnimationExporters/GIFAnimationExporter.gd
```diff
@@ -12,15 +13,15 @@ func _init():
...
 	progress_report_method,
 	progress_report_args
 ) -> PoolByteArray:
-	var exporter = GIFExporter.new(images[0].get_width(), images[0].get_height())
-	for i in range(images.size()):
-		exporter.add_frame(images[i], durations[i], MedianCutQuantization)
+	var f = frames[0]
+	var exporter = GIFExporter.new(f.content.get_width(), f.content.get_height())
+	for v in frames:
+		exporter.add_frame(f.content, f.duration, MedianCutQuantization)
```
OverloadedOrama, line 25: I think this line should be `v.content` and `v.duration` instead of `f.content` and `f.duration`. It seems to be saving only the first frame over and over in the exported gif.

20kdc: that is indeed correct, whoops

20kdc: Fixed as of 6709172de3b8d5771473ffc6ff794a120efaabde

OverloadedOrama (approved):
Looks good to me. Great job, thank you!

PR #798, opened 2022-12-23
Selection modes

Variable-ind:
1) Removed HSeparators from tool headers
2) Added Selection modes:

- 4 modes Default (that obeys shortcuts)
- Other 3 don't obey shortcuts and will lock a specific mode (handy when we need to do adding over and over, and so on)
### Preview:

https://user-images.githubusercontent.com/77773850/209374031-9d453b27-a5cc-4463-990c-4a3b7134b318.mp4

### DIFF
--- a/src/Tools/BaseTool.gd
+++ b/src/Tools/BaseTool.gd
@@ -19,7 +19,7 @@ func _ready() -> void:
 		color_rect.color = Global.left_tool_color
 	else:
 		color_rect.color = Global.right_tool_color
-	$Header/Label.text = Tools.tools[name].display_name
+	$Label.text = Tools.tools[name].display_name
 
 	load_config()
 
--- a/src/Tools/BaseTool.tscn
+++ b/src/Tools/BaseTool.tscn
@@ -15,26 +15,11 @@ margin_right = 116.0
 margin_bottom = 4.0
 rect_min_size = Vector2( 0, 4 )
 
-[node name="Header" type="HBoxContainer" parent="."]
+[node name="Label" type="Label" parent="."]
 margin_top = 8.0
 margin_right = 116.0
 margin_bottom = 22.0
-custom_constants/separation = 0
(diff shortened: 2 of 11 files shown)


OverloadedOrama:
I'm testing it and overall it works okay, but there is some inconsistency regarding the shortcuts. When "Always Add" is selected and <kbd>Control</kbd> is being held, it's subtracting, and intersecting with <kbd>Shift</kbd>+<kbd>Control</kbd>, which I think should be the desirable behavior. But when "Always Subtract" is selected, it ignores the add and intersect shortcuts. Similarly, "Always Intersect" seems to ignore the add shortcut, but not the subtract shortcut.

I think it should behave like in GIMP, where holding a shortcut should always change the mode, and have it reflect on the UI as well. So, for example, if "Always Subtract" is selected but the user holds <kbd>Shift</kbd> to add, the mode should change to "Always Add" (the OptionButton should change as well), and return to "Always Subtract" when <kbd>Shift</kbd> is released. Perhaps we should also use icons instead of an OptionButton, like GIMP does, but this can change after the PR is merged.

image
Screenshot from GIMP.

Variable-ind:
> I'm testing it and overall it works okay, but there is some inconsistency regarding the shortcuts. When "Always Add" is selected and Control is being held, it's subtracting, and intersecting with Shift+Control, which I think should be the desirable behavior. But when "Always Subtract" is selected, it ignores the add and intersect shortcuts. Similarly, "Always Intersect" seems to ignore the add shortcut, but not the subtract shortcut.

Okay i resolved the shortcut conflicts :+1:

> (the OptionButton should change as well)

but i didn't know of a way to do this so i left it for you :smiley:  (though personally i don't think this is a good idea because there also exists the "Default" option.)

OverloadedOrama (approved):
Alright, looks good to me. Thank you!

PR #800, opened 2022-12-28
Allow floating point for Horizontal and Vertical Rulers

Variable-ind:
Used `stepify()` function to restrict numbering up-to one decimal place

### DIFF
--- a/src/UI/Canvas/Rulers/HorizontalRuler.gd
+++ b/src/UI/Canvas/Rulers/HorizontalRuler.gd
@@ -76,8 +76,10 @@ func _draw() -> void:
 			)
 			var val = (ruler_transform * major_subdivide * minor_subdivide).xform(Vector2(j, 0)).x
 			draw_string(
-				font, Vector2(position.x + RULER_WIDTH + 2, font.get_height() - 4), str(stepify(val, 0.1))
-			)
+				font,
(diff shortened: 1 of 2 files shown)


src/UI/Canvas/Rulers/HorizontalRuler.gd
```diff
@@ -76,7 +76,7 @@ func _draw() -> void:
 			)
 			var val = (ruler_transform * major_subdivide * minor_subdivide).xform(Vector2(j, 0)).x
 			draw_string(
-				font, Vector2(position.x + RULER_WIDTH + 2, font.get_height() - 4), str(int(val))
+				font, Vector2(position.x + RULER_WIDTH + 2, font.get_height() - 4), str(stepify(val, 0.1))
```
OverloadedOrama, line 79: What is the difference between using `stepify` and simply removing `int`? I'm noticing the same result using both ways, either only whole numbers, or .5 numbers.

Variable-ind: not much differencebat the moment,
i noticed the floats became very long numbers (e.g 2.00001) etc if i change the sub-divisision variables
in short, using stepify is a more "General" solution

OverloadedOrama (approved):
Thank you!

PR #802, opened 2023-01-04
Api Improvements

Variable-ind:
1. Tools are now properly Removed
2. A new `get_current_cel_info()` method
3. A new `show_error()` method. (useful for displaying messages like "Incompatible API" etc...)


OverloadedOrama (approved):
Looks good to me, thank you!

PR #804, opened 2023-01-05
Folders rearrangement

Variable-ind:
Organized the UI folder into sub folders
### Before
image
### After
image


src/Main.tscn
```diff
@@ -64,6 +64,7 @@ __meta__ = {
 [node name="CreateNewImage" parent="Dialogs" instance=ExtResource( 28 )]
 
 [node name="OpenSprite" parent="Dialogs" instance=ExtResource( 12 )]
+rect_min_size = Vector2( 180, 63 )
```
OverloadedOrama, line 67: Is there a reason for all these `rect_min_size` changes? They seem unrelated to this PR, and if they are to change, they should change in their own scene file rather than in Main.tscn.

src/UI/Nodes/TransparentChecker.tscn
```diff
@@ -0,0 +1,75 @@
...
+shader_param/color1 = Color( 0.7, 0.7, 0.7, 1 )
+shader_param/color2 = Color( 1, 1, 1, 1 )
+shader_param/offset = Vector2( 0, 0 )
+shader_param/scale = Vector2( 0, 0 )
+shader_param/rect_size = Vector2( 0, 0 )
+shader_param/follow_movement = false
+shader_param/follow_scale = false
+
+[sub_resource type="GDScript" id=2]
+script/source = "extends ColorRect
```
OverloadedOrama, line 18: Shouldn't this code be in TransparentChecker.gd instead of being embedded in TransparentChecker.tscn? It wasn't this way before.

Variable-ind: All these changes were done by godot itself in the process of moving
I'll try to change them to how they were before

Variable-ind: Done

src/Main.tscn
Variable-ind: Done

OverloadedOrama (approved):
Okay, everything seems to be good now. Thank you!

PR #805, opened 2023-01-05
Updated Tools Folder Arrangement

CybearWarfare:
Seeing as other folders were being cleaned up and organized I thought it would be beneficial to arrange the tools folder and give it some more structure. Tools have now been moved into respective sub folders with the exception of base tool, and tool button.


OverloadedOrama:
I like the idea, but I think a better structure would be to put similar tools together in the same folder, like selection tools currently are, maybe based on the class they extend from. For example, Pencil, Eraser, Shading and the shape tools could be under the same `DrawTools` folder. And in that folder, the shape tools could be placed in a `ShapeTools` subfolder. I think that would be enough as it is, but if more categorization is wanted, then Pan and Zoom could be under a `CanvasTools` subfolder. Bucket could also fit under `DrawTools`, but since it doesn't extend from Draw.gd I wouldn't recommend it.

CybearWarfare:
> 

Updated to your specifications.

PR #806, opened 2023-01-06
Added the Perspective editor

Variable-ind:
This was one of my extensions, i massively improved the original code, and added it to the core so **Heavy Testing Encouraged**
### A Demonstration of Power:

https://user-images.githubusercontent.com/77773850/210975490-efeb55c0-59e4-494c-b43d-3a09eb21bd97.mp4


luiq54:
Looks great! :smile: 
This is my preliminary review: 
1) Don't use a colored background for element like "Point 1". It makes a bit harder to read. Instead you can use a small icon next to it or use a small filled square. 
2) Don't use colored foreground for buttons. It is distracting (red on dark bg is difficult to read). Instead use a undo point when deleting and prompt user with a warning dialog if it is something major

one more smaller change. change `=` in X and Y to a colon (:) to maintain consistency with other panels

Variable-ind:
> Looks great! smile This is my preliminary review:
> 
>     1. Don't use a colored background for element like "Point 1". It makes a bit harder to read. Instead you can use a small icon next to it or use a small filled square.
> 
>     2. Don't use colored foreground for buttons. It is distracting (red on dark bg is difficult to read). Instead use a undo point when deleting and prompt user with a warning dialog if it is something major
> 
> 
> one more smaller change. change `=` in X and Y to a colon (:) to maintain consistency with other panels

How about this?
Screenshot from 2023-01-06 15-21-28

luiq54:
Looks better! We should probably add icons to buttons throughout the editor (Little side project for future :smile: )

image
It will be better if these line elements were added vertically instead of horizontally. 
Also can these elements be made into collapsible widgets like Godot editor instead of popups ? 
image

And move the delete button to the very bottom so that its more prominent and people like me don't misclick it when trying to change color :sweat_smile:

OverloadedOrama (changes requested):
image
There seems to be an issue with the Line collapsible containers. Expanding the Line1 container in the screenshot, creates another empty collapsible container directly below it, which doesn't seem to do anything. This also happens for every line container, but not in any of the other collapsible containers, like "Lines" or "Point: 1".

I also think that the perspective editor should probably be next to the canvas, but hidden by default, as most people will probably not need it. If you're having difficulties changing the Layout resources, I can do it for you.

Apart from that, I'm not 100% sure about the current status of the UI, but we can make further changes after the PR is merged.

Variable-ind:
Done

> image There seems to be an issue with the Line collapsible containers. Expanding the Line1 container in the screenshot, creates another empty collapsible container directly below it, which doesn't seem to do anything. This also happens for every line container, but not in any of the other collapsible containers, like "Lines" or "Point: 1".
> 
> I also think that the perspective editor should probably be next to the canvas, but hidden by default, as most people will probably not need it. If you're having difficulties changing the Layout resources, I can do it for you.
> 
> Apart from that, I'm not 100% sure about the current status of the UI, but we can make further changes after the PR is merged.

Done

OverloadedOrama (approved):
Fantastic work, thank you very much!

PR #807, opened 2023-01-08
Added supported_api_versions parameter

Variable-ind:
This will implement a new parameter called `supported_api_versions`
Note that the extension will still work without this parameter in `extension.json` but if it is included, then pixelorama will do an additional check if it's current api is present in the `supported_api_versions`

an example of extension.json containing this parameter is given below (this extension will now only work on pixelorama Api versions 1 and 2). So in case Api version 3 contains breaking changes it will simply show an error dialog instead of a BIG LOUD CRASH
```swift
{
"author": "Variable",
"description": "Test the new Api",
"display_name": "TestAPI",
"license": "MIT",
"name": "TestAPI",
"nodes": [ "Main.tscn" ],
"version": "0.1",
"supported_api_versions": [1, 2]
}
```


OverloadedOrama (approved):
Looks good, thanks!

PR #811, opened 2023-01-14
Fix previews

Variable-ind:
At-least i managed to fix the move tool preview

### DIFF
--- a/src/UI/Canvas/Canvas.gd
+++ b/src/UI/Canvas/Canvas.gd
@@ -51,7 +51,9 @@ func _draw() -> void:
 						selected_layers.append(cel_pos[1])
 			if i in selected_layers:
 				if Global.current_project.layers[i].can_layer_get_drawn():
-					draw_texture(current_cels[i].image_texture, move_preview_location, modulate_color)
+					draw_texture(
+						current_cels[i].image_texture, move_preview_location, modulate_color
+					)
 				else:
 					draw_texture(current_cels[i].image_texture, Vector2.ZERO, modulate_color)
 			else:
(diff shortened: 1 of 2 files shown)


src/UI/Canvas/Canvas.gd
```diff
@@ -45,7 +45,12 @@ func _draw() -> void:
 			continue
 		var modulate_color := Color(1, 1, 1, current_cels[i].opacity)
 		if Global.current_project.layers[i].is_visible_in_hierarchy():
-			if i == current_layer:
+			var selected_layers = []
+			if move_preview_location != Vector2.ZERO:
+				for cel_pos in Global.current_project.selected_cels:
+					if cel_pos[0] == Global.current_project.current_frame:
```
OverloadedOrama, line 51: I think an extra check is needed in case one of the layers is locked or invisible and its content does not get moved, with `can_layer_get_drawn()`. Otherwise it looks good to me I think.

Variable-ind: Done

src/UI/Canvas/Canvas.gd
```diff
@@ -45,8 +44,18 @@ func _draw() -> void:
 			continue
 		var modulate_color := Color(1, 1, 1, current_cels[i].opacity)
 		if Global.current_project.layers[i].is_visible_in_hierarchy():
-			if i == current_layer:
-				draw_texture(current_cels[i].image_texture, move_preview_location, modulate_color)
+			var selected_layers = []
+			if move_preview_location != Vector2.ZERO:
+				for cel_pos in Global.current_project.selected_cels:
+					if cel_pos[0] == Global.current_project.current_frame:
```
OverloadedOrama, line 50: Wouldn't it work if the `can_layer_get_drawn()` check was added here, so unavailable layers wouldn't be added in the `selected_layers` array at all?

Variable-ind: Done

OverloadedOrama (approved):
Thank you!

PR #813, opened 2023-01-18
Implement a spacing option for the pencil tool

Variable-ind:
Added Snap mode to the pencil tool that can allow you to do dithering easily

https://www.youtube.com/watch?v=lxYXghPFYlk&feature=youtu.be


OverloadedOrama:
This looks great, but I'm not too sure about the name Snap mode. Perhaps changing it to "Brush spacing" or plain "Spacing" might be better. That would also be consistent with Krita's spacing brush setting.

OverloadedOrama (approved):
Looks good to me, thank you!

PR #816, opened 2023-01-20
Make the color picker pick any visible color on the canvas, regardless of layer

Variable-ind:
Color Pickers will now pick the top color by default


OverloadedOrama (approved):
My only concern is that if a project has a lot of layers, it might be slow to loop through all of these layers to find the color. But I cannot think of faster approach right now, and I agree that this behavior should be the default, so it should be good to merge. Thank you!

PR #817, opened 2023-01-20
cel checker visibility based on it's content

Variable-ind:


Variable-ind:
A better system is now implemented

PR #818, opened 2023-01-21
Make separate window for small canvas

Variable-ind:
This add the ability of (Panel <-----> Window)
small canvas uses this feature at the moment but others can use it as well


Variable-ind:
i'll remove this pull for now (It was a fun experiment though)

PR #833, opened 2023-03-12
Set tilemask automatically

Variable-ind:
This will automatically set a tilemask whenever we change project or cell,
No need for setting it manually
This also removed the need for it to be stored in a Project

### DIFF
--- a/src/Main.tscn
+++ b/src/Main.tscn
@@ -64,6 +64,7 @@ __meta__ = {
 [node name="CreateNewImage" parent="Dialogs" instance=ExtResource( 28 )]
 
 [node name="OpenSprite" parent="Dialogs" instance=ExtResource( 12 )]
+rect_min_size = Vector2( 170, 59.5 )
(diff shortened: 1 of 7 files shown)


src/Main.tscn
```diff
@@ -64,6 +64,7 @@ __meta__ = {
 [node name="CreateNewImage" parent="Dialogs" instance=ExtResource( 28 )]
 
 [node name="OpenSprite" parent="Dialogs" instance=ExtResource( 12 )]
+rect_min_size = Vector2( 170, 59.5 )
```
OverloadedOrama, line 67: I'm guessing this is just Godot adding unnecessary changes to Main.tscn again? :P

Variable-ind: Hmmm, i think so (i thought i prevented main from getting committed 😅) feel free tho remove them 😅

OverloadedOrama (approved):
I think this is a nice change. Thank you!

PR #834, opened 2023-03-12
TileMode for selection, (plus some other fixes)

Variable-ind:
1. A compact version of tile-mode for selections
2. Fixed some bugs with the ORIGINAL tile mode (Image brushes needed to be adjusted)

https://user-images.githubusercontent.com/77773850/224560348-ac7f0663-59c2-4b44-81d0-a8279a949428.mp4


OverloadedOrama:
In that case maybe we should keep it in the Selection menu, but make it a checkbox. If in the future there is a need to group different selection behaviors, we could create a Behavior sub-menu then.

OverloadedOrama (approved):
Thanks!

PR #835, opened 2023-03-15
Spritesheet Animation Canvas Preview

Variable-ind:
even though pixelorama is designed to make sprite-sheets frame by frame instead of all in one frame
BUT...
still many still make entire sprite-sheets on a single frame/image (and complain about animating afterwards LOL)
This will now be possible

https://user-images.githubusercontent.com/77773850/225286066-b24bf367-435f-4719-a095-4bb2f32d6ac7.mp4


OverloadedOrama (approved):
Thanks!

PR #839, opened 2023-03-26
Vector layers

mrtripie:
- Currently experimental
- Plan is to have a text tool implemented for the first version, new shapes and tools to edit them can be added later


OverloadedOrama:
An idea I had that could simplify vector layers a lot was to wait for Godot 4.2 and use `Image.load_svg_from_string()`. We could store all vector data in SVG format inside the vector cels, then compose a string and draw the vector shapes using that method. This should work well, although I can see two potential issues:
- We will still need to handle text differently, as Godot can't load svg images that have text, unless they're being rendered as paths.
- We would need to re-generate the entire SVG string every time a vector shape changes, which could potentially be slow for large images or for cels that contain a lot of vector data.

Issue #842, opened 2023-04-06
mirrored drawing line(s) is not centered by default


ShadowFungi:
**Pixelorama version: 0.10.3**

**OS/device including version: ArchLinux - 6.2.9-arch1-1**

**Issue description: Mirrored/symmetrical drawing line(s) is not centered, thus causing the right pixels to be cut off when drawing too far to the left, or the left side not touching the edging if drawing to the right.**

**Steps to reproduce:**

1. create a 32x32 document
2. enable horizontal mirrored drawing (left/right mirror)
3. draw from "centered" line to right edge of canvas
4. notice that a pixel is missing on the left side

PR #843, opened 2023-04-09
fixed mirrored drawing

ChrisTs8920:
Issue #842.


OverloadedOrama (commented):
Thank you for the contribution! I don't think this is the correct approach, as this offsets the mirrored pixels by one pixel, which makes the symmetry look wrong. What issue #842 is referring to is the default placement of the symmetry axes. I think simply removing "+ 0.5" from lines 73, 74, 79 and 80 in Project.gd should be a better solution.

PR #848, opened 2023-04-22
Update .desktop file and add new MimeInfo

luiq54:
Pixelorama's .desktop file now contains all the file types it can open by default (Used by DEs to provide app suggestions in context menus).
Also adds a MimeInfo file for `.pxo`. It provides context for the format and can be used by packages to register and associate the file format with the application (like flatpak)


OverloadedOrama (approved):
Thank you!

PR #851, opened 2023-04-30
Signal api

Variable-ind:
Added a SignalsAPI to easily connect and disconect important signals like:
`project_changed`
`cel_changed`
`color_changed`
`texture_changed` (this one gets emitted whenever current cel's texture is changed. it also auto-adjusts to a new "current cel" if project or cel is selected, useful for extensions like voxelorama that may need to hide/show depth indicators as soon as a pixel is drawn/erased)

Added a new method `get_cel_info_at()` to ProjectAPI


OverloadedOrama (approved):
Thanks!

PR #854, opened 2023-05-06
Fix top color picking

Variable-ind:
The "Current layer" mode was able to pick-colors from a locked layer, while the "Top Color" mode was not;
this will fix that issue (i only needed to see if layer was visible or not)


OverloadedOrama (approved):
Thanks!

PR #857, opened 2023-05-10
[Targeted for 0.11.0] Fix Onion skin settings don't save between runs

### LINKED ISSUE #757: Onion skin settings don't save between runs
**Pixelorama version:**
 v0.10-2 stable

 Windows 10
I put Pixelorama in a folder synced with Mega cloud drive, idk if it's related, I know Windows 10 can be pretty glitchy about cloud sync folders.

**Issue description:**
Changing onion settings does not save them after closing and restarting. Which is annoying, as I usually don't want to see future frames or see them 'above canvas' so I have to change them every time.

**Steps to reproduce:**
Change onion skin settings, restart program. Onion skin setting return to default (1 frame future, 1 frame back, both above canvas).

Variable-ind:
Fixes #757


OverloadedOrama (approved):
Looks good, thank you!

PR #858, opened 2023-05-13
[Targeted for 0.11.0] Changes to PanelAPI

Variable-ind:
1. Added tab is now set by layout instead of specifying an "alongside_node"
2. Added tab is now shown in the Panels Sub menu
3. Yields are no longer needed in extension code, instead it is carried out by ExtensionAPI itself

**Extension code for testing**
```swift
extends Node

var tab
# This script acts as a setup for the extension
func _enter_tree() -> void:
	# add a test panel as a tab
	tab = Panel.new()
	tab.name = "test"
	ExtensionsApi.panel.add_node_as_tab(tab)

func _exit_tree() -> void:
	ExtensionsApi.panel.remove_node_from_tab(tab)
```

### DIFF
--- a/src/Autoload/ExtensionsAPI.gd
+++ b/src/Autoload/ExtensionsAPI.gd
@@ -2,14 +2,14 @@
 extends Node
 
 # use these variables in your extension to access the api
-var general = GeneralAPI.new()
-var menu = MenuAPI.new()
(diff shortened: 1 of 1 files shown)


OverloadedOrama (approved):
Thanks!

PR #861, opened 2023-05-16
[Continuation] Auto select cloned frames, reuse tag animations

Variable-ind:
Continuation of #648,
in the original pull request i wanted that on cloning, the newly cloned frames are all selected by default so that if one wishes to clone animations (as shown below) then this would make it easier but i gave up half way cause i couldn't figure out how...

with my recent attempt, i've successfully implemented this feature
### Applications:
1) easily create animations with twice the speed

https://github.com/Orama-Interactive/Pixelorama/assets/77773850/59b2ef7f-a335-4fba-91f9-066e97236322

2) Another useful application of this is to easily create animations similar to the one in the video below (if you analyze the video, you will notice many instances where a single animation is played again and again with an offset which as a whole makes it beautiful) https://youtu.be/Kh-KrYun8SY


Variable-ind:
with the cels now being selected automatically, i also took the liberty of adding the ability to re-use already existing tag animations as you see below

https://github.com/Orama-Interactive/Pixelorama/assets/77773850/4a650b83-b803-4fe6-9423-44922640e56c

OverloadedOrama (approved):
Looks good to me, thanks!

PR #862, opened 2023-05-16
[Targeted for 0.11.0] Make Onion skin update live with the "Move" tool

Variable-ind:
While using the "Move" tool, the onion skin will move it's past and future frames accordingly if they are in selected cels to give a better idea of where it will land,
(NOTE: selections will still work as it did previously i-e update onion shin after transformation is confirmed)

https://github.com/Orama-Interactive/Pixelorama/assets/77773850/e82b180b-c175-4c54-8e7c-ceada8ca1fd6


OverloadedOrama (approved):
Looks good to me, thank you!

PR #865, opened 2023-05-19
Update opacity across selected frames

20kdc:
This causes the opacity slider to update all selected frames of the current layer rather than just the current frame.
This is useful because opacity is treated independently of cel linking, so without an option like this it's hard to work with layers you want to have one global opacity.

An example of this is if you have, say, a "noise layer" on your sprite (i.e. import noise from some other application, cut it to fit the sprite, then adjust opacity to taste). This is useful for adding a bit of texture. But combining a noise layer with animation is awkward because the opacity is per-frame, so you have to have everything setup at the start or individually adjust each layer.


OverloadedOrama (approved):
Looks good to me, thanks!

PR #866, opened 2023-05-19
Limit onion skinning slider max values to 10

Variable-ind:
Limited value sliders of past and future frames to 10 and enabled the `allow greater` option so you can still set it to values greater than 10


OverloadedOrama (approved):
Thank you!

Issue #697, opened 2022-05-08
GIFs are sometimes exported as solid color


novhack:
**Pixelorama version:**
0.10-stable

**OS/device including version:**
EndeavourOS 5.17.5, Mesa 22.0.3, Radeon RX 5700 XT
MrTriPie was able to reproduce the issue on Windows 10

**Issue description:**
When exporting images with more than 256 colors Pixelorama sometimes exports a GIF which is just a solid color.
Exporting images which don't require quantization it seems to work properly. I expect the issue is somewhere in that code.

**Steps to reproduce:**
This standard test image of Lenna causes the issue sometimes:
![Lenna_(test_image)](https://user-images.githubusercontent.com/42614907/167296039-6e4753aa-9654-4b40-86c4-5efa5583e6a2.png)

Variable-ind:
I can confirm this on Linux with the master branch, but can't seem to produce it on v0.8.1 and below (this needs further confirmation)

OverloadedOrama:
I think this may due to multi-threaded rendering not working properly in Godot 3.x. In Godot 4 it outright crashes with the compatibility renderer, so I disabled it. I tested gif exporting with multi-threaded rendering with the Forward+ as well, and in both cases this bug doesn't seem to be longer reproducible in Pixelorama v1.0, so I'll go ahead and close the issue now.

PR #868, opened 2023-05-23
Export Gif Message

Variable-ind:
adds a message instructing user what to do in case Gif export goes wrong

Screenshot from 2023-05-23 23-13-40


OverloadedOrama (approved):
Thanks! We should keep in mind to remove the message once #697 is solved.

PR #869, opened 2023-05-24
[Targeted for 0.11.0] Export options to extension

Variable-ind:
Continuation of #810
A Looooong awaited feature to be added to the ExtensionsApi
You can now add custom export options using extensions. but wait!!!, there's more.... they can also contain gdnative!

The main reason why gdnative wasn't being loaded was that the libraries (the ones that end with with .so, .dll,  etc...), can't be loaded from a virtual file system (e.g .pck) so what the workaround is to simply place the libraries in a github repo and download them through HttpRequest (it can be a one-time process)

So now that i've figured out the issue and created an example extension (which was the condition in the original PR)

https://github.com/Orama-Interactive/Pixelorama/assets/77773850/1f05f564-bc06-4600-a6f9-c74b192961e2

### DIFF
--- a/src/Autoload/Export.gd
+++ b/src/Autoload/Export.gd
@@ -218,6 +218,7 @@ func export_processed_images(
 			var result := true
 			var details := {
 				"processed_images": processed_images,
+				"durations": durations,
 				"export_dialog": export_dialog,
 				"export_paths": export_paths,
 				"project": project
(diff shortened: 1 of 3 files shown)


Variable-ind:
Here is an updated extension (it's a full implementation of Godout)
[Godout.zip](https://github.com/Orama-Interactive/Pixelorama/files/11567751/Godout.zip)

Here's the pck file (remember to unzip it first)
[Godout.pck.zip](https://github.com/Orama-Interactive/Pixelorama/files/11567757/Godout.pck.zip)

OverloadedOrama (approved):
Thanks!

PR #876, opened 2023-06-13
Fix Ok button Text of Save and Export Dialog

Variable-ind:
to "Save" (also changes to "Save & Quit" when needed) and "Export"


OverloadedOrama (approved):
Thanks!

PR #836, opened 2023-03-18
Animate image effect properties

Variable-ind:
A foundation to be improved upon,
this sets a very basic animation to some image effect properties (interpolation from ~~0~~ an initial value to a final value)
~~the downside is that ranges can't be set like (rotation from 60 to 70), outline from (5 to 10px) etc...~~

https://user-images.githubusercontent.com/77773850/226129860-3dc4bb0e-a5bd-4700-90b1-7fb3c873fce2.mp4

https://user-images.githubusercontent.com/77773850/226130143-d6c0f3a3-b6dc-4db7-b9fc-b75551f0736f.mp4

PR #879, opened 2023-06-19
ImageEffect Animation 2.0

Variable-ind:
2nd iteration of #836 (also fixed 2 bugs in previous iteration)
moved the animation system to it's own separate `AnimatePanel.tscn`

## Added:
- It is now possible to preview other frames as well, not just the current frame

## Changed:
- Made the animate system more intuitive
- Property animation now uses SceneTreeTween instead of Lerp which allows us to set:
  - Transition type and Ease mode
- Animation starts from left of timeline and ends at right (in old system animation just went through whatever order cels were placed in selected cel array)

## Fixed:
- [0.11.0 and earlier issue] fixed issue when changing `affect` mode to something else and the back to `Selected cels` made the animate system appear even if no animatable properties were set for that image effect
- [0.11.0 and earlier issue] fixed value and saturation sliders being divided by 360 (when they should be divided by 100). this bug was introduced in #836 (sorry about that)

## Visuals:
here's animating an outline from 10 to 20 in old system

https://github.com/Orama-Interactive/Pixelorama/assets/77773850/fba2570c-e9c3-444a-9e77-20017d362504

here's doing the same thing in the new system

https://github.com/Orama-Interactive/Pixelorama/assets/77773850/3e0b97f2-edbe-46e4-ab38-8ac8fe856327


OverloadedOrama (approved):
This is amazing, excellent work. Thank you!

PR #880, opened 2023-06-20
Set alignment of the LineEdit Node for file name in Export Dialog to Left

Hamster5295:
The LineEdit node for file name in Export Dialog seems to behave weirdly when the Align property is set to Right, as is shown in the gif:

before

In the picture, I pressed Backspace once per time, but at the second time, 2 charactors disappeared together with the caret.
The cause of the bug remains uncleared, but simply setting the Align property to Left solves the problem like this:

after

## Changes
* The LineEdit for file name in Export Dialog now aligns on the left side


Hamster5295:
Now uses gifs instead

OverloadedOrama (approved):
Thanks!

PR #884, opened 2023-07-04
Move Centralize Image to the frame menu

Variable-ind:
1. Moved the Centralize from "Image Effects" to "Frame Menu"
2. And make it effect on Selected frames instead of "Single Frame"


OverloadedOrama (approved):
Looks good, thank you!

PR #893, opened 2023-08-02
Spritesheet Autoslicer

Variable-ind:
https://github.com/Orama-Interactive/Pixelorama/assets/77773850/b54c5e47-407e-444d-bdad-46f959861819

Implementation of a basic auto slicer

Maintainer edit: This also fixes a bug where importing a large image as a new project and then importing a smaller image as a new frame or layer, would cause errors when attempting to draw outside of the smaller image boundaries.


OverloadedOrama (approved):
This is a very useful feature, thank you!

PR #894, opened 2023-08-04
Added integer zoom

Variable-ind:
this is now an option in preferences>zoom as seen below
Screenshot from 2023-08-04 20-27-26

it will restrict the zoom to only integral multiples of 100% (e.g 200%, 300%, ....)

https://github.com/Orama-Interactive/Pixelorama/assets/77773850/e6fad966-b315-4ed7-9e76-bfcb678fb3c6


src/Preferences/PreferencesDialog.tscn
```diff
@@ -401,33 +400,52 @@ margin_right = 486.0
...
 text = "On"
 
+[node name="Label2" type="Label" parent="HSplitContainer/VBoxContainer/ScrollContainer/RightSide/Canvas/ZoomOptions"]
+margin_left = 306.0
+margin_top = 5.0
+margin_right = 486.0
+margin_bottom = 19.0
+rect_min_size = Vector2( 110, 0 )
+size_flags_horizontal = 3
+text = "Integer Zoom (1x, 2x, 3x, ...)"
```
OverloadedOrama, line 434: It might be best if this is just "Integer Zoom", as its functionality is already explained in the hint tooltip.

Variable-ind: Done

OverloadedOrama (approved):
Looks good to me, thank you!

PR #896, opened 2023-08-06
Fix a bug in integer zoom

Variable-ind:
1. Do nothing further if zoom margin is infinite
2. Fit to Frame now obeys integer zoom


OverloadedOrama (approved):
Thanks!

PR #898, opened 2023-08-13
Add a silhouette option to reference images

TheLsbt:
This pull request adds a option to make images a silhouette. <br>
This is done by darkening all the opaque pixels. <br>

This is the UI changes: <br>
Screenshot (40)

### DIFF
--- a/src/Preferences/PreferencesDialog.tscn
+++ b/src/Preferences/PreferencesDialog.tscn
@@ -11,14 +11,14 @@
 [sub_resource type="ButtonGroup" id=1]
 
 [node name="PreferencesDialog" type="AcceptDialog"]
-visible = true
-margin_right = 571.0
-margin_bottom = 441.0
+margin_right = 620.0
+margin_bottom = 450.0
 window_title = "Preferences"
 resizable = true
 script = ExtResource( 1 )
 __meta__ = {
 "_edit_horizontal_guides_": [  ],
+"_edit_use_anchors_": false,
 "_edit_vertical_guides_": [  ]
 }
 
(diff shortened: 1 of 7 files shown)


TheLsbt:
I think I fixed the linter issue.

src/UI/ReferenceImages/ReferenceImage.gd
```diff
@@ -63,6 +65,14 @@ func deserialize(d: Dictionary) -> void:
 			# don't do FLAG_REPEAT - it could cause visual issues
 			itex.create_from_image(img, Texture.FLAG_MIPMAPS)
 			texture = itex
+		# Apply the silhouette shader
+		var mat = ShaderMaterial.new()
+		mat.shader = preload("res://src/Shaders/SilhouetteShader.gdshader")
```
OverloadedOrama, line 70: It's generally better to preload stuff in the variable declaration space and then use that variable later, for example by creating a variable called shader and setting it to this `var shader := preload("res://src/Shaders/SilhouetteShader.gdshader")` at the top of the code and then in this line you could do `mat.shader = shader`.

src/Preferences/PreferencesDialog.tscn
```diff
@@ -11,14 +11,14 @@
 [sub_resource type="ButtonGroup" id=1]
 
 [node name="PreferencesDialog" type="AcceptDialog"]
-margin_right = 620.0
-margin_bottom = 450.0
+visible = true
```
OverloadedOrama, line 14: Looks like you might've forgotten to set it back to invisible 😅. From what I see, the changes in PreferencesDialog.tscn are just margin value changes that Godot makes automatically and don't really need to be included, unless you've manually made some changes to the nodes yourself. If you haven't, it's safe to just discard the changes made in this file completely.

TheLsbt:
I believe that the issues with the Preference Dialog should be fixed (The file doesn't appear in **Files Changed**).  
I also created a variable to assign the shader to in ```RefrenceImage.gd```.

OverloadedOrama (approved):
Looks good to me, thank you!

PR #900, opened 2023-08-15
Port to Godot 4

OverloadedOrama:
This pull request is upgrading Pixelorama from Godot 3.5.2 to Godot 4.1. It should be working properly but more testing is required. The 3.x codebase can be found in the [0.x branch](https://github.com/Orama-Interactive/Pixelorama/tree/0.x).

New features/improvements:
- Overall speedups, most noticeably in the bucket fill algorithm. The below speeds are measured inside the Godot editor, release builds should be faster. Tested with Godot 3's GLES2 renderer and Godot 4's compatibility renderer.

| WxH | 3.x | 4.x|
| - | - | - |
| 1024x1024 | 1705 ms | 732 ms |
| 2048x2048 | 6722 ms | 2664 ms |
| 4096x4096 | 26502 ms | 11044 ms |
- The brush increment/decrement shortcuts can now be changed.
- The font size can be changed from the Preferences, separately from the display scale.
- Every shader-based image effect is automatically working without the need to change renderers, and they all work now on the Web version. This comes at the cost of less compatibility, as the desktop version now requires OpenGL 3.3 minimum instead of 2.1, and the Web version requires WebGL 2 instead of WebGL 1.
- Added missing shortcuts for buttons. TODO: Add more default shortcuts, if they are needed. Implements the final part of #588.
- 3D layers now support torus shapes.
- Image effect animation now supports the tweening transition method of [spring](https://github.com/godotengine/godot/pull/76899).
- ~10000 less lines of code! Although most of them are related to themes and scenes.

Bug fixes:
- Dividing by zero in value sliders and spinboxes no longer crashes the program.

Known issues:
- ~~Holding Control + Z does not keep undoing (same for redo). Needs either https://github.com/godotengine/godot/pull/64317 or https://github.com/godotengine/godot/pull/36493 to be merged.~~ EDIT: Done, waiting for Godot 4.2 now.
- ~~Clicking outside the area of popup windows no longer closes them~~ EDIT: Fixed
- ~~The new global shortcuts for brush size incrementing and decrementing cannot be changed yet, as Keychain needs to be updated to support Godot 4's Input Map abilities, such as setting keyboard modifiers along with mouse buttons.~~ EDIT: Fixed
- ~~GIF exporting does not always work~~ EDIT: Fixed, but gif exporting is no longer using multi-threading because it does not work well yet with the compatibility renderer (https://github.com/godotengine/godot/issues/61650).
- ~~Cannot load images in the Web version.~~ EDIT: Fixed by https://github.com/godotengine/godot/pull/81015, waiting for Godot 4.1.2.
- The application takes more time to open, around 1-1.5 seconds more than it did with Godot 3.x after some optimizations, on my machine.
- Certain UI elements do not look well in some themes
- The size of the exported binaries has increased due to Godot 4's new features. This could be mitigated with https://docs.godotengine.org/en/stable/contributing/development/compiling/optimizing_for_size.html. However, the .pck file is now smaller in size because the themes have been optimized and the DroidSansFallback font has been deleted, as system fonts can now be used as fallback.
- Extensions can't add extra file types
- APNGs do not seem to be imported in the Godot project, which temporarily no longer allows for animated splash screens
- Cannot change the renderer yet, although this may not needed, especially if https://github.com/godotengine/godot-proposals/issues/7379 gets implemented.

Compatibility breaking issues:
- Extensions made for v0.x will no longer work on v1.x.
- Projects with 3d cels from v0.x may not be imported properly or at all.
- Shortcut profiles from v0.x will not work on v1.x.

Stuff not tested yet:
- ~~CI~~ EDIT: Seems to be working now.
- ~~Other platforms besides Linux~~ EDIT: Also tested Windows and Web.
- ~~Extensions~~ EDIT: Updated [Voxelorama](https://github.com/Orama-Interactive/VoxeloramaExtension) to Godot 4, seems to be working well.
- ARM exporting (for devices such as the Raspberry Pi)


theofficialgman:
> ARM exporting (for devices such as the Raspberry Pi)

@OverloadedOrama just an fyi, I don't know how many people actually use your github releases for "pi".
At pi-apps for arm32 and arm64 (1000 pixelorama users) I build godot from source since the "pi" godot releases you use can't be used due to too high and inconsistent glibc requirements, in addition to poor choice of tuning variables.

theofficialgman:
> Unfortunately I do not own a Raspberry Pi 4 or any ARM device that runs Linux to test the builds myself.

Well ARM hardware is everywhere (likely your phone, Nintendo Switch, lots of embedded devices) so I find it hard to believe you wouldn't have any you could readily use.

Regardless, I typically build on an always free (its really free) ARM Ampere server from Oracle https://www.oracle.com/cloud/free/ even though I have a lot of ARM hardware available to me. I have my free oracle ARM server setup with 200GB of storage, 24GB of ram, 4 cores, running Ubuntu Bionic ARM64 with an ARMhf chroot for any 32bit ARM builds I might need to do.

Compiling Godot 4 is no different than any other platform. Following the official documentation works fine.

For example this is how I built my export templates on the ARM64 Bionic server (using GCC-8 since 7 currently has issues building Godot-4 https://github.com/godotengine/godot/issues/79352 )

```bash
scons platform=linuxbsd target=template_release production=yes lto=full bits=64 arch=arm64 CC=gcc-8 CPP=cpp-8 CXX=g++-8 FC=gfortran-8
```
and this gives me `godot.linuxbsd.template_release.arm64`

OverloadedOrama:
The only ARM device I own is my phone, which already runs Android so I'm not sure how I can test Linux ARM builds on it. I didn't know about the Oracle ARM servers though, I will definitely take a look. Thanks for letting me know!

PR #904, opened 2023-08-24
remove old (pre 0.8) .pxo loading code

Variable-ind:


OverloadedOrama:
Some context for this decision. Versions earlier than 0.8 used a different system for .pxo files that changed in version 0.8, which came out back in September 23 2020, almost 3 years ago. Most people use newer versions than this, so it makes little sense to keep the old compatibility code anymore. Removing this code also allows us to properly detect if a pxo that is being loaded is corrupt or not. Prior to this pull request, loading a corrupt pxo file results in a crash.

OverloadedOrama (approved):
Thanks!

PR #906, opened 2023-08-29
fix: mute some divide int warning

ppphp:
godot 4 works terrible on this warning.


ppphp:
I would fix gdtoolkit first

PR #911, opened 2023-09-06
Layer blend modes

OverloadedOrama:
Alternative to #688, much more straightforward way to handle layer blending now that we use Godot 4, but I would really appreciate feedback as I'm not sure if this is the best way to handle layer blending and if it has any disadvantages over any other methods. CC @AlphinAlbukhari @mrtripie.

I also didn't encounter any layer limit with this method, I tested up to 300 layers and it seemed to be working perfectly with no performance issue. I did have to add a 1024 layer limit though in order to pass opacities and blend modes for each layer as arrays to the shader, since shaders don't support uniform arrays with unknown sizes. I am worried that 1024 might be too high for certain GPUs, so we may have to lower the number even more. Alternatively, we could either pass opacities and blend modes as textures, where each pixel would represent a layer, or we could change the shader code during runtime to update the size of the array, although I am worried this might affect performance as we would re-generate the shader every time we add or remove a layer.

image

Current blending modes supported (thanks to #688 and https://godotshaders.com/snippet/blending-modes/):
- Normal
- Darken
- Multiply
- Color burn
- Linear burn
- Lighten
- Screen
- Color dodge
- Additive (linear dodge)
- Overlay
- Soft Light
- Hard Light
- Difference
- Exclusion
- Subtract
- Divide
- Hue
- Saturation
- Color
- Luminosity

This PR only works as a preview, I would like feedback first to ensure that this method is indeed a good one before I continue the work.

To-do:
- [x] Preview blending in the canvas.
- [x] Take cel opacity into account.
- [x] Change blend mode in the UI.
- [x] Preview blending in tile mode.
- [x] Preview blending in the image effect dialogs.
- [x] Make blending work on exported images.
- [x] Save blend mode in pxo files.
- [x] Make merging layers take blending into account.
- [x] Preview blending in the other canvases.
- [x] Show the move tool preview.
- [x] Undo/redo for when the blending mode changes.

Can be done after this PR is merged:
- Optimize it so the Texture2DArray doesn't get re-created for every change, only when the number of layers or the project changes.
- Support layer group blending.


OverloadedOrama:
I think I will merge this PR as it is so it will get tested more as part of the master branch, and potential issues can be fixed later. For now, it should be stable enough.

PR #914, opened 2023-10-08
Allow user to move Palettes to Trash/Recycle Bin instead of deleting them permanently

Variable-ind:
Some additional changes that I thought be added to this pull request.

1. Fix deleted signal emitted wrongly
   - it was written as `deleted.emit(deleted)` which did not seem to be working so i changed it to `deleted.emit()`

2. Fix create and edit palette dialog UI
    - Dialog titles of `CreatePaletteDialog.tscn` and `EditPaletteDialog.tscn`
    - Label Autowrap and alignment to match how they were in v0.11.2 (In `CreatePaletteDialog.tscn` and `EditPaletteDialog.tscn`)
    - a minor ui change in `EditPaletteDialog.tscn` that allows comment to expand vertically

3. fix path nodes renamed to path3d by godot project converter
    - `ReferenceImageButton.tscn` and `EditPaletteDialog.tscn`

PR #918, opened 2023-10-09
[0.x] Allow user to move Palettes to Trash/Recycle Bin instead of deleting them permanently 

Variable-ind:
0.x version of #914.


OverloadedOrama (approved):
Thanks!

PR #919, opened 2023-10-09
Ask user if they wish to delete Extensions permanently or not

Variable-ind:
- move the faulty extension to a separate folder
- don't delete incompatible extensions (v0.11.x --> v1.0). so that they can still be used with (0.11.x)
- on uninstall ask user if they wish extension to move to bin or permanent deletion


OverloadedOrama (approved):
Thanks!

PR #920, opened 2023-10-11
If "Save & Exit" is pressed while quitting, don't close the software if the project failed to save

Variable-ind:


OverloadedOrama (approved):
Looks good to me, thanks!

PR #921, opened 2023-10-11
[0.x] If "Save & Exit" is pressed while quitting, don't close the software if the project failed to save

Variable-ind:
0.x version of #920


OverloadedOrama (approved):
Thanks!

PR #928, opened 2023-10-25
[0.x] Fix cel3d compatibility when transferring project from 1.0 to 0.x

Variable-ind:
3d cels should be backwards compatible now as well in case user decides to switch back to 0.x for it's lighting


OverloadedOrama (approved):
Thanks!

PR #929, opened 2023-10-26
Improve cel3d Gizmos

Variable-ind:
- The axis arrow will disappear if it comes very close to the origin to make the gizmos more cleaner.
- The axis arrows will always remain ahead of the scale circle
- If both arrow and scale icon end up being in the same place then the scale function will take priority

https://github.com/Orama-Interactive/Pixelorama/assets/77773850/cc327196-c78c-4da4-a432-eecb7f8ab825


OverloadedOrama (approved):
Thanks!

PR #940, opened 2023-11-11
Implement layer effects

OverloadedOrama:
Introduces a non-destructive workflow for effects/filters as proposed in #619.

https://github.com/Orama-Interactive/Pixelorama/assets/35376950/4cd36686-6aa8-48fb-a575-dfa64bfbeb61
https://github.com/Orama-Interactive/Pixelorama/assets/35376950/9b4a03be-4ce0-48e1-a8df-a422d5df98f8
(These are videos, not sure why they aren't embedded)

All it does is essentially takes a copy of the images of the cels, uses `ShaderImageEffect` to apply each shader of every effect into the copy, and display that copy to the canvas (if Display Layer Effects is enabled) and to the exported image. The image data of the cels are not affected in any way, thus making it non-destructive.

The biggest issue with this system is performance. Drawing is noticeably laggier when effects are enabled, and this is due to `RenderingServer.force_draw()` that `ShaderImageEffect` is using. This method is necessary however, without it the resulting image is blank. A better solution may be this https://github.com/godotengine/godot/pull/75436, or to use something like this https://github.com/godotengine/godot-proposals/issues/7379.

The UI is different than the one proposed in #619, because I used a newer mockup by mrtripie which was posted on our Discord server. Feel free to give feedback!

To avoid code duplication with the destructive image effect UI and to future-proof the system, the parameter UI of the effects is being generated on the fly by reading the shader code and making the appropriate UI elements for each uniform (ValueSliders for ints and floats, CheckBoxes for bools, ColorPickerButtons for vec4s, etc). This system is future-proofed in the sense that we 1) don't have to update the UI every time a shader changes (although we have to do it anyway for the destructive image effects as of right now), and 2) it supports custom shaders loaded by the user, whenever we implement this feature.

Features not yet implemented, not inside this PR's scope:
- Save the effect data on pxo, as well as the result images for easier importing (by third-party software) and exporting. Probably not going to work on it since the pxo format is subject to change, see #939.
- Support for animation, let properties change between frames and interpolate between them.
- Let users load shaders. I was thinking that we could have a "Shaders" folder next to the "Brushes", "Palettes" etc folders where users can put their own `.gdshader` files, and for every shader file we would add both a destructive effect in the Image menu, and a non-destructive layer effect.
~~- A way to permanently apply effects.~~ EDIT: Implemented in cc8d18cea63a348353b85599f6fd17ef7ea761fe.


OverloadedOrama:
Should be finished for now. I just implemented undo/redo support, as well as a way to permanently apply effects (destructively) to all cels of the layer. The main thing missing is pxo saving, but this can be handled after we deal with #939.

PR #941, opened 2023-11-16
[0.x] Make cloned frames only select a cel if it's coresponding original cel was selected as well

Variable-ind:
0.x version of #942


OverloadedOrama (approved):
Thanks!

PR #943, opened 2023-11-19
fixed some documentation typos

Variable-ind:
also fixed 2 functions of signal api using old (target, method) arguments instead of callable


OverloadedOrama (approved):
Thanks!

PR #945, opened 2023-11-19
Added a way to select cels through API using select_cels()

Variable-ind:
Plus some ProjectAPI revisions


OverloadedOrama (approved):
Thanks!

PR #946, opened 2023-11-22
Improved the Paste tag system

Variable-ind:
this is the second iteration of paste tag system
it will now allow user to paste tags across projects
needs to be thoroughly tested
## Rules

1. Pasting system will take frames from layers of the "from project", and paste them to current project, inside corresponding layers of the same name.
2. In case a layer of same name doesn't exist in the current project, a new layer will be created for it.


src/UI/Timeline/PasteTagPopup.gd
```diff
@@ -1,37 +1,237 @@
...
-	tag_container.gui_input.connect(_on_TagContainer_gui_input)
+	var tag_container: Control = Global.animation_timeline.find_child("TagContainer")
+	# connect signals
+	tag_container.connect("gui_input", _on_TagContainer_gui_input)
+	from_project_list.connect("item_selected", _on_FromProject_changed)
+	animation_tags_list.connect("item_selected", _on_TagList_id_pressed)
+	create_tags.connect("toggled", _on_CreateTags_toggled)
+
+
+func refresh_list() -> void:
```
OverloadedOrama, line 21: I think `animation_tags_list.clear()` should be here instead of `_on_TagContainer_gui_input()`. Currently changing the projects from the project list OptionButton is still showing the tags of the previously selected project, instead of the list being cleared and re-made.

Variable-ind: done :+1:

OverloadedOrama (approved):
Thanks!

PR #947, opened 2023-11-23
Update pl_PL.po

piotrtrkostrzewski94:
Corrections to the translation of the Polish version


OverloadedOrama:
Hello, thank you for the contribution. Translations are only being accepted through Pixelorama's [Crowdin page](https://crowdin.com/project/pixelorama) and not through pull requests, so please apply your changes there and they will eventually get into Pixelorama.

PR #957, opened 2023-12-10
[ExtensionsAPI] Implement an Import API

Variable-ind:
Adds an import api that can allow you to:
override pixelorama's import images dialog to use the image for a different purpose that the default (e.g in video i used it to open image in a new panel for no reason), but you can also do other things like change how a sprite-sheet gets imported etc...

https://github.com/Orama-Interactive/Pixelorama/assets/77773850/6659c7c3-f316-4355-8a78-e253157a7e85

## Other related addition
the import dialog couldn't get any of the extension's node by itself so i added another api method

```
func get_main_nodes(extension_name: StringName) -> Array[Node]:
```
(Returns the initial nodes of an extension named `extension_name`.
initial nodes are the nodes whose paths are in the `nodes` key of an extension.json file.)

Here's the test extension
Code:
[TestExtension (Code).zip](https://github.com/Orama-Interactive/Pixelorama/files/13627561/TestExtension.Code.zip)

Extension:
[TestExtension(Extension).zip](https://github.com/Orama-Interactive/Pixelorama/files/13627562/TestExtension.Extension.zip)


OverloadedOrama (approved):
Thanks!

PR #963, opened 2023-12-15
Allow Internal extensions

Variable-ind:
Okay... this is a big one...

Internal extensions are extensions that are not **.pck** files, but actual code (arranged in the same format as you would an extension) in `res://src/Extensions/`

this can allow for potentially making pixelorama modular, As a test i made the tools into an extension (only a test, so it's not part of this pr)

https://github.com/Orama-Interactive/Pixelorama/assets/77773850/8efd5570-a727-4aeb-9091-faa98df80f64

these extensions will not appear in extension list automatically so you would have to use `_add_internal_extensions` method located in `HandleExtensions.gd`

Screenshot from 2023-12-15 23-10-31


Variable-ind:
An addition change for this pr:
The tools folder contents have been re-arranged
Screenshot from 2023-12-20 18-13-35

OverloadedOrama (approved):
Looks good, thank you!

PR #970, opened 2024-01-08
Make the "Add Object" button more obvious that it is a button.

TheLsbt:
Change the menu button to flat to indicate that it is a button that you can press and not a label.


OverloadedOrama (approved):
Eventually we may want to replace these text buttons with icons, but this is good for now. Thank you!

TheLsbt:
@OverloadedOrama that would be a better idea

PR #971, opened 2024-01-08
Better signal representation

TheLsbt:
Changed a few signals in Global.gd, namely `project_changed` and `cel_changed` to better represent what they actually do.
Re-added `project_changed` but this time it get's emitted when a project is changed for example when it the canvas is drawn on.

## **Note**
I did not change the methods related in `ExtensionsAPI.gd` because I don't want to break any extensions. I did change actual signals that get called but not the functions.


OverloadedOrama (approved):
Looks good! As for the method names in `ExtensionsAPI.gd`, I think they should be changed as well to avoid confusion. Since 1.x is meant to be a breaking change, we shouldn't worry about compatibility that much right now. Extensions made in for 0.x will not work in 1.x anyway due to the switch to Godot 4. I will merge this for now and we could look at the ExtensionsAPI method names later. Thank you!

PR #972, opened 2024-01-10
 Polish context menu - pixelorama.nsi

piotrtrkostrzewski94:
I added a translation to the Edit project with Pixelorama context menu into Polish.

### DIFF
--- a/installer/pixelorama.nsi
+++ b/installer/pixelorama.nsi
@@ -6,7 +6,7 @@
 ; Helper variables so that we don't change 20 instances of the version for every update
 
   !define APPNAME "Pixelorama"
-  !define APPVERSION "v0.11.2"
+  !define APPVERSION "v0.11.3"
   !define COMPANYNAME "Orama Interactive"
 
 
@@ -81,7 +81,7 @@
 
 ; Multilingual support
 
-  !insertmacro MUI_LANGUAGE "English"
+  !insertmacro MUI_LANGUAGE "Polski"
   ;@INSERT_TRANSLATIONS@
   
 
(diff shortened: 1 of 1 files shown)


mystchonky:
Translations are done through crowdin on https://crowdin.com/project/pixelorama. You can submit translations there and they get imported into the project every few weeks. It's easier for translators to work on it this way.

piotrtrkostrzewski94:
> Translations are done through crowdin on https://crowdin.com/project/pixelorama. You can submit translations there and they get imported into the project every few weeks. It's easier for translators to work on it this way.
I know this, I'm a Pixelorama translator there.  There is no context menu added on CrowdIn.  It's not a problem for me to add a commit here.  I hope the creators don't mind this.

mystchonky:
Well in that case, that entry needs to be added to crowdin instead :)

mystchonky:
Never mind this is for installer and not the main application. I have no clue how those are handled :sweat_smile:

piotrtrkostrzewski94:
Maybe they will add it for other languages, there is no need for Polish :) In my opinion, it is for the main application because it is in the context menu :)

installer/pixelorama.nsi
```diff
@@ -81,7 +81,7 @@
 
 ; Multilingual support
 
-  !insertmacro MUI_LANGUAGE "English"
+  !insertmacro MUI_LANGUAGE "Polski"
```
OverloadedOrama, line 84: This change looks incorrect to me, it seems to be replacing English with Polish rather than adding it as a new translation. I'm not entirely familiar with how .nsi files work so I'm not sure what the correct solution is here.

installer/pixelorama.nsi
```diff
@@ -152,7 +152,7 @@
       WriteRegStr HKCR "Pixelorama project\DefaultIcon" "" "$INSTDIR\pxo.ico"
 
       WriteRegStr HKCR "Pixelorama project\shell\open\command" "" '$INSTDIR\${APPNAME}.exe "%1"'
-      WriteRegStr HKCR "Pixelorama project\shell\edit" "" "Edit project in ${APPNAME}"
```
OverloadedOrama, line 155: Same as above.

piotrtrkostrzewski94: In the file I sent, there is a Polish version of the context menu on line 155.  Should I add a new file
Screenshot_2024-01-11-15-38-07-831_com android chrome
 called pixelorama-pl.nsl?

piotrtrkostrzewski94: I made a mistake, I could have added it immediately as a new file and added PR.  Sorry, I'm still learning Github.  I sent the correct (I hope) PR.

PR #973, opened 2024-01-11
Create pixelorama pl.nsi

piotrtrkostrzewski94:
Polish translation of the Pixelorama installer and context menu.


OverloadedOrama:
I don't think this pull request is going to work either, just adding a new file is not going to have any effect. To be honest, I have no idea how nsi files are supposed to be translated. I suspect only one .nsi file would be needed with multiple strings for each language, but I'm not sure if that is actually true as I am unfamiliar with how .nsi files work. We should figure this out first before creating any new PRs.

piotrtrkostrzewski94:
I think that you could add a languages option in the installer and add a given nsi file to the appropriate language

PR #978, opened 2024-01-19
Moved live preview code to ImageEffect Dialog

Variable-ind:
live/delayed preview now works with all effects, to allow this feature, simply change the visibility of `LiveSettings` node in the target effect dialog.


OverloadedOrama (approved):
Thanks!

PR #988, opened 2024-02-29
Add flip X and Flip Y, and Rotation (90, 180, 270) buttons to BaseTool

RorotoSic:
This will be really useful for custom brushes. (rather than creating 8 different brushes, it will be just one) ^^

image


RorotoSic:
I think it's always best to reduce the number of clicks as much as possible, especially in art software, to keep the user in the creative flow.
I also thought about putting buttons on the tool's global options, but they will aplly modifications on all tools like "pixel perfect" do and its better to have each tools their own modificiations.
(Personaly in krita i only use the "Edit brushes" to CREAT new brush , never to update one during painting process, but it can be just me.)


* True and correction made
* It's quicker to update brush when flip/rotate are above. and i plan to code an option to keep brush folder open to preview flip and rotate modifcation in live to unlock creative posibility by seeing them that we may not thinked about. Can i sugest to add an expand button to keep the ui clean (as i made in the gif) ?
* It's true, same result but more clicks and a little brain work ^^ it's always easier to do with a dedicated button
*I modified the radio buttons and I don't know if I like it. Previously the rotations were applied on the brush we see in the preview (that brush may already had some flip and rotate changes), now with the radio button our brain has to remember the original brush (the one without changes ), remove the modification activated that we see on preview and then apply the desired change. Both approaches works, i dont know wich one is better.


ezgif-1-659ce0a639

OverloadedOrama:
I think the expand button works fine. We may need to reconsider the UI in the future if we add more options, but for now I think it should be fine. By the way, we have a custom `CollapsibleContainer` node (which can be added by adding a new node in the Scene tree and searching "CollapsibleContainer"), and you can put the elements that can be collapsed/expanded as children of that node, without needing to add any extra code for hiding/unhiding. You can also add text to the CollapsibleContainer node, perhaps something like "Rotation options".

As for the radio button, I don't really have a strong opinion. I feel like it makes more sense this way, but I'm not totally against your original implementation, so feel free to choose which implementation you prefer.

RorotoSic:
I think as radio buttons seems to be more common we can keep them.
here is the current state :
https://github.com/Orama-Interactive/Pixelorama/assets/70805756/f2767771-af66-4944-be64-8576ecfdb197

src/Autoload/Tools.gd
```diff
@@ -1,6 +1,7 @@
 extends Node
 
 signal color_changed(color, button)
+signal flip_rotate(flip_x, flip_y, rotate_90, rotate_180, rotate_270)
```
OverloadedOrama, line 4: The [GDScript style guide recommends using past tense for signal names](https://docs.godotengine.org/en/latest/tutorials/scripting/gdscript/gdscript_styleguide.html#signals), so this should be `flip_rotated`.

src/Tools/BaseDraw.tscn
```diff
@@ -25,7 +41,54 @@ anti_aliasing = false
 [node name="ToolOptions" instance=ExtResource("2")]
 script = ExtResource("3")
 
-[node name="Brush" type="HBoxContainer" parent="." index="2"]
+[node name="RotationOptions" type="VBoxContainer" parent="." index="2"]
+visible = false
+layout_mode = 2
+theme = SubResource("Theme_i7q3e")
```
OverloadedOrama, line 47: I think adding an entire theme just to remove the CollapsibleContainer's internal button decoration might be a bit overkill. Buttons have a `flat` property that does exactly that. [In the latest commit on the master branch](https://github.com/Orama-Interactive/Pixelorama/commit/6f3796627f7e93e8ed53660928546cee0b1cd7ad), I added a `flat` property in CollapsibleContainer that updates the `flat` status of its internal button. I think this could be used instead of a theme.

OverloadedOrama (commented):
Okay, changes in the UI look good now I think! Left some nitpick comments about the code, and it should be good to get merged after that.

RorotoSic:
"signal flip_rotated(flip_x, flip_y, rotate_90, rotate_180, rotate_270) " was not t a conflict before , i dont know why sudenlly it is while nothing has changed in my code

OverloadedOrama:
Nothing to worry about, it's because I made a change to the master branch and it happened to affect that specific line. I fixed the conflict now.

OverloadedOrama (approved):
Should be good to get merged now. Excellent work, thank you very much!

PR #989, opened 2024-03-03
Add a "Share brush config" buttons to BaseTool

RorotoSic:
Painting and erasing with the same brush feels really good. (It's a bit like Krita for painting and erasing).
Painting with the same brush but with different colors is also a very nice feeling (I know we can change colors with x but it takes two hands to do it... and having to manually change also the size and the rotation/flip is creative flow killer)
This new feature it’s an accessory that keeps artist in the creative flow ! 

ezgif-7-a8c2b176af

### DIFF
--- a/src/Autoload/Tools.gd
+++ b/src/Autoload/Tools.gd
@@ -1,6 +1,7 @@
 extends Node
 
 signal color_changed(color, button)
+signal flip_rotate(flip_x, flip_y, rotate_90, rotate_180, rotate_270)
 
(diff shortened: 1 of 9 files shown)


RorotoSic:
I'm thinking that is a button that can go in the global tool option 🤔

OverloadedOrama:
> I'm thinking that is a button that can go in the global tool option 🤔

Yeah I think it would be better as a global tool option as well.

OverloadedOrama:
Thank you for the contribution! Superseded by #1120.

Issue #791, opened 2022-12-06
Mac: Built-in palettes not found


zellyn:
**Pixelorama version:**
0.10.3, both .dmg and homebrew

**OS/device including version:**
Mac M1 Max

**Issue description:**
The builtin palettes are not found in the UI

**Steps to reproduce:**
Install Pixelorama. Look at the dropdown in the Palettes menu. There are no palettes.

**Reason**

If you create a new palette, it is saved in `/Applications/Pixelorama.app/Contents/MacOS/pixelorama_data/Palettes`.
The included palettes are stored in `/Applications/Pixelorama.app/Contents/Resources/pixelorama_data/Palettes`.
If you copy the included palettes from `Resources` to `Contents/MacOS`, they show up in the UI.

ivanquirino:
Same issue on Pixelorama version 0.11.3, mas OS 12.3, intel cpu

haythamnikolaidis:
I am experiencing the same problem, MacOS 13.1, intel CPU, version 0.11.4 a quick check and it looks like on a fresh install the `pixelorama_data` folder is not found in the in the `Contents/MacOS` folder at all this means, that the default patterns, brushes and palettes do not load.

My initial attempt to work around this, I downloaded the `pixelorama_data` folder from the main branch, to replace the missing folder, (they get created as empty on first run) this resolved the problem for the brushes and the patterns, but not the palettes, after stumbling upon the work around in this thread I realised that the palettes files in the main branch are `json` while the ones shipped with Pixelorama are `tres.` Replacing the `json` files with the `tres` ones resolves the problem for palettes.

PR #1008, opened 2024-04-19
Make default palettes available for clean installs on MacOS

haythamnikolaidis:
This PR aims to resolve : [791](https://github.com/Orama-Interactive/Pixelorama/issues/791)

On MacOS with a clean install the application is unable to reference the `pixelorama_data` directory in the `Pixelorama.app/Contents/Resources` directory, this meant that `Palettes.gd` was unable to copy the default palette files to the user directory.

The above was cause by 3 issues: 
1. `root_directory` in `Globals.dg's _init()` function was set to `.` which resolved to the directory of the binary (`Pixelorama.app/Contents/macOS/`)
2. The `is_instance_valid()` seemed to always return false in `_get_palette_files()` in `Palettes.gd` while running as an exported application (it worked well in the editor.)
3. Even with the `pixelorama_data` directory being copied to the `Pixelorama.app/Contents/Resources` directory, `DirAccess` would fail to `open` the Resources directory.

The solution I have implemented is to set the `root_directory` to "res://" for MacOS, so that it maps to the virtual directory of the `Pixelorama.pck` file. 
Check the validity of `dir` in `_get_palette_files()` with a null check that should cover both of the previous conditions, eliminating the need for `is_instance_valid()`.
Then (probably controversially) removed the `.gdignore` file from the `pixelorama_data` directory before export of the release builds, this ensure that the `pixelorama_data` directory is packed into the `Pixelorama.pck` file and available at run time.

I was unsure why the `.gdignore` file was there to begin with, but I was hesitant to remove it all together incase it interfered with the other platforms, therefore I thought removing it in the automated builds made the most sense, to keep this as a change that purely affects MacOS.

### DIFF
--- a/src/Autoload/Global.gd
+++ b/src/Autoload/Global.gd
@@ -660,7 +660,7 @@ func _init() -> void:
 	if OS.has_feature("template"):
 		root_directory = OS.get_executable_path().get_base_dir()
 	if OS.get_name() == "macOS":
-		root_directory = "res://"
+		data_directories.append(root_directory.path_join("../Resources").path_join(CONFIG_SUBDIR_NAME))
 	data_directories.append(root_directory.path_join(CONFIG_SUBDIR_NAME))
 	if OS.get_name() in ["Linux", "FreeBSD", "NetBSD", "OpenBSD", "BSD"]:
 		# Checks the list of files var, and processes them.
--- a/src/Autoload/Global.gd
+++ b/src/Autoload/Global.gd
@@ -660,7 +660,9 @@ func _init() -> void:
 	if OS.has_feature("template"):
 		root_directory = OS.get_executable_path().get_base_dir()
 	if OS.get_name() == "macOS":
-		data_directories.append(root_directory.path_join("../Resources").path_join(CONFIG_SUBDIR_NAME))
+		data_directories.append(
+			root_directory.path_join("../Resources").path_join(CONFIG_SUBDIR_NAME)
+		)
 	data_directories.append(root_directory.path_join(CONFIG_SUBDIR_NAME))
 	if OS.get_name() in ["Linux", "FreeBSD", "NetBSD", "OpenBSD", "BSD"]:
 		# Checks the list of files var, and processes them.
--- a/src/Autoload/Global.gd
+++ b/src/Autoload/Global.gd
@@ -659,6 +659,8 @@ func _init() -> void:
 	loaded_locales.sort()  # Make sure locales are always sorted
 	if OS.has_feature("template"):
 		root_directory = OS.get_executable_path().get_base_dir()
+	if OS.get_name() == "macOS":
+		root_directory = "res://"
 	data_directories.append(root_directory.path_join(CONFIG_SUBDIR_NAME))
 	if OS.get_name() in ["Linux", "FreeBSD", "NetBSD", "OpenBSD", "BSD"]:
 		# Checks the list of files var, and processes them.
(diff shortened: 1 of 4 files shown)


src/Autoload/Global.gd
```diff
@@ -659,6 +659,8 @@ func _init() -> void:
 	loaded_locales.sort()  # Make sure locales are always sorted
 	if OS.has_feature("template"):
 		root_directory = OS.get_executable_path().get_base_dir()
+	if OS.get_name() == "macOS":
+		root_directory = "res://"
```
OverloadedOrama, line 663: The `root_directory` variable is used in other places as well, such as when saving the `override.cfg` file which is used to change certain project settings (like `embed_subwindows` and `per_pixel_transparency`). For this to work, `override.cfg` has to be in the same folder as the executable, so I think changing this to "res://" is going to break that functionality, although I could be mistaken, unfortunately I do not own a macOS device to test it.

The reason palettes, brushes and patterns are not bundled inside the application is to allow users to remove or edit the default ones if they don't like them, as well as add new ones. Will this still be possible with this PR?

haythamnikolaidis: Thank you. I did not notice `override.cfg` I will have to find a work around for `root_directory`. 

With regards to the palettes, they remain editable (they are copied to the `user://` on start up) but they can not be deleted (if a default palette is deleted, on next start up the palette is restored) which I don't think is functionality we want. However I do think that is a separate bug because if the `pixelorama_data` folder was found in the `Pixelorama.app/Content/Resources/` this would still be the case.

Regardless, I'll find a work-a-round for the `root_directory` then over the weekend I can compare the functionality on my Widows machine and make sure that MacOS functions similarly.

haythamnikolaidis:
I reverted the changes to the workflows as well as the `root_directory` variable. It now resolves finding the `pixelorama_data` directory by adding the `Pixelorama.app/Content/Resources` directory to the `data_directories` list.

This resolves the original problem, without affecting `override.cfg` or needing to change the automated builds.

This leaves me with a new concerns though:

On start up the app will copy the default palettes to the `/Library/Application Support/pixelorama/` (`user://`) directory. The code always checks to see if the default palettes are in the `user://` directory, if they are not it copies them over again. This means renaming the `default` palette to anything else will recreate it as `default`, leaving the user with two default palettes. I replicated this on both MacOS and windows, however it seems to only affect the main branch as it does not happen in `0.77.4` Should we create a new issue for this?

OverloadedOrama:
> I reverted the changes to the workflows as well as the `root_directory` variable. It now resolves finding the `pixelorama_data` directory by adding the `Pixelorama.app/Content/Resources` directory to the `data_directories` list.

Awesome, thank you!

> On start up the app will copy the default palettes to the `/Library/Application Support/pixelorama/` (`user://`) directory. The code always checks to see if the default palettes are in the `user://` directory, if they are not it copies them over again. This means renaming the `default` palette to anything else will recreate it as `default`, leaving the user with two default palettes. I replicated this on both MacOS and windows, however it seems to only affect the main branch as it does not happen in `0.77.4` Should we create a new issue for this?

Yeah, I think a new issue should be created for this. Since it doesn't happen on macOS only, it's a general issue and it should be unrelated to this PR and #791. If this PR is ready to get merged, I think I should go ahead and merge it now and we can fix that other issue later.

haythamnikolaidis:
I am happy that we merge this. I am also happy to create the issue to track the new default palettes bug.

OverloadedOrama (approved):
Thank you!

PR #1018, opened 2024-05-10
Return relationship lines to Tree theme

Kiisu-Master:
I noticed that the tree relationship lines were visible in the Godot 3.x ver so i added them back.
It's not exactly the same because i also enabled the effect where it makes lines width bigger when selecting something.

I configured all themes the same way except light theme where i made the line darken instead of brighten. It wouldn't be visible otherwise.

This change is only visible in the shortcuts editor:

image


OverloadedOrama (approved):
Good catch, thank you!

PR #1019, opened 2024-05-16
Curve tool implementation

Variable-ind:
(side note) This Pr's code should be checked in case i've left some redundant code in there.

https://github.com/Orama-Interactive/Pixelorama/assets/77773850/dccfba41-96cf-496e-83a4-805a96f879b3


OverloadedOrama (approved):
Looks great to me, thank you very much!

PR #1025, opened 2024-06-08
Implement Columns field of imported .gpl palettes.

Anaminus:
- Fixes Columns field being ignored when imported.
- `_fill_imported_palette_with_colors` is modified to accept an optional width parameter, which defaults to Palette.DEFAULT_WIDTH instead of a hard-coded 8.
- A maximum of width of 16384 is enforced to mitigate maliciously large values.

### DIFF
--- a/src/Autoload/Palettes.gd
+++ b/src/Autoload/Palettes.gd
@@ -525,8 +525,10 @@ func _import_image_palette(path: String, image: Image) -> Palette:
 
 	return _fill_imported_palette_with_colors(path.get_basename().get_file(), colors)
 
+
 ## Maximum allowed width of imported palettes.
 const MAX_IMPORT_PAL_WIDTH = 1 << 14
+
 
 ## Fills a new [Palette] with colors. Used when importing files. Dimensions are
 ## determined by taking colors as a one-dimensional array that is wrapped by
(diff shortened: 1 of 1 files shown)


src/Autoload/Palettes.gd
```diff
@@ -464,8 +467,11 @@ func _import_gpl(path: String, text: String) -> Palette:
 		elif line.begins_with("Name: "):
 			palette_name = line.replace("Name: ", "")
 		elif line.begins_with("Columns: "):
-			# Number of colors in this palette. Unnecessary and often wrong
-			continue
+			# The width of the palette.
+			line = line.trim_prefix("Columns: ")
```
OverloadedOrama, line 471: On my tests, it doesn't seem to work because the `is_valid_int()` check returns false. There's probably some whitespace in the remaining string that makes the condition fail. I managed to fix this by replacing this line with

`line = line.trim_prefix("Columns: ").strip_edges()`

Anaminus: I agree. It looks like GIMP does the same thing by using g_strstrip.

OverloadedOrama (approved):
Looks good to me, thank you!

PR #1034, opened 2024-07-11
Fixes to perspective editor

Variable-ind:
- removed some garbage code.
- made tracker toggle obey the variable in the script (on startup).
- Changed type of Perspective editor from `Control` to `PanelContainer` to avoid some issues with dockable container


OverloadedOrama (approved):
Thank you!

PR #1038, opened 2024-07-19
changed `Open Logs Folder` to `Open Editor Data Folder`

Variable-ind:
this seems to be how Godot does it, plus this is a more general approach that also helps those that are looking to copy their layouts, and settings, over to some other computer.

other change in this pr includes:
- fixed temporary directory for downloading extensions gets created in the wrong place


OverloadedOrama (approved):
Thanks!

PR #1043, opened 2024-07-24
Fix certain strings not updating when language changes

nicejammer:
Currently, a few buttons and labels in Pixelorama aren't getting updated when changing languages, they just remain at the language the program started up with or only update when changing projects.

For the example below: I set the language to Greek, restarted the program and then switched the language back to English.
image
Other than these buttons in the File menu, there are different places where this also happens (see the changes included).

Here's how this PR fixes it:
- For the simple strings, all that was needed to do was remove the `tr()` when setting them up as they need to stay in English for `TranslationServer` to work with them.
- For the strings that have formatting (that is, strings that can only be partially translated, like the Save button above which has the project's name in it), `TranslationServer` cannot do the automatic translations, so it was needed to add code to retranslate them manually whenever the language is changed.

**Note:** The issue is still present in `FileDialog`s, but in that case it's a problem with Godot and it seems like it was already fixed in 4.3.

### DIFF
--- a/src/Palette/EditPaletteDialog.gd
+++ b/src/Palette/EditPaletteDialog.gd
@@ -32,9 +32,9 @@ var old_name := ""
 func _ready() -> void:
 	export_file_dialog.use_native_dialog = Global.use_native_file_dialogs
 	# Add delete and export buttons to edit palette dialog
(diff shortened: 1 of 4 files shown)


OverloadedOrama (approved):
Looks good to me, thank you very much!

PR #1044, opened 2024-07-25
added palette docs, allow more than one dimentional ValueSlider, add …

Variable-ind:
…more error messages (it seems `push_error` is a more efficient way than `printerr`)


OverloadedOrama (approved):
Thank you!

PR #1046, opened 2024-07-27
update SmartSlicer

Variable-ind:
This pr just updates the SmartSlicer to a newer version, nothing new is exposed to the import options,

though user may experience a slight boost in performance.


OverloadedOrama (approved):
Thanks!

PR #1047, opened 2024-07-27
fix some things in ExtensionExplorer

Variable-ind:
- fix wrong texture flag of large preview
- add fail safe if extension in version in repo file is read as a string
- add failsafe if the string is not a link


OverloadedOrama (approved):
Thanks!

PR #1051, opened 2024-07-28
Add missing translation strings and fix broken ones

nicejammer:
Adds many missing strings to `Translations.pot` and fixes a few ones that were broken either because of missing periods, differences in capitalization or because of trailing newlines (the latter seems to get automatically trimmed in tooltips during runtime, better watch out for those in the future).
I don't know if anything is lost when an existing source string is changed in crowdin, so I mostly updated the broken strings in Pixelorama instead where I could in order to match the ones in the POT.


OverloadedOrama (approved):
Good catch, thank you!

PR #1056, opened 2024-07-30
Add setting for Godots low processor usage mode

Kiisu-Master:
Someone reported problems with lagging UI on discord. It looks like it is Godots low processor usage mode not working correctly, so i thought we could add a setting for disabling it.

### DIFF
--- a/src/Autoload/Global.gd
+++ b/src/Autoload/Global.gd
@@ -475,7 +475,7 @@ var fps_limit := 0:
 ## Found in Preferences. Affects the per_pixel_transparency project setting.
 ## If [code]true[/code], it allows for the window to be transparent.
 ## This affects performance, so keep it [code]false[/code] if you don't need it.
-var update_continuously := true:
+var update_continuously := false:
 	set(value):
 		update_continuously = value
 		OS.low_processor_usage_mode = !value
--- a/src/Autoload/Global.gd
+++ b/src/Autoload/Global.gd
@@ -475,12 +475,10 @@ var fps_limit := 0:
 ## Found in Preferences. Affects the per_pixel_transparency project setting.
 ## If [code]true[/code], it allows for the window to be transparent.
 ## This affects performance, so keep it [code]false[/code] if you don't need it.
-var low_processor_usage := true:
+var update_continuously := true:
 	set(value):
-		if value == low_processor_usage:
-			return
-		low_processor_usage = value
-		OS.low_processor_usage_mode = value
+		update_continuously = value
+		OS.low_processor_usage_mode = !value
 var window_transparency := false:
 	set(value):
 		if value == window_transparency:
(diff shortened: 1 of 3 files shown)


src/Autoload/Global.gd
```diff
@@ -680,6 +686,7 @@ func _init() -> void:
 	if ProjectSettings.get_setting("display/window/tablet_driver") == "winink":
 		tablet_driver = 1
 	single_window_mode = ProjectSettings.get_setting("display/window/subwindows/embed_subwindows")
+	low_processor_usage = ProjectSettings.get_setting("application/run/low_processor_mode")
```
OverloadedOrama, line 689: I'm not sure if this line is needed, as the low processor usage mode is being changed with `OS.low_processor_usage_mode` and not as a project setting.

src/Preferences/PreferencesDialog.tscn
```diff
@@ -1058,6 +1058,19 @@ mouse_default_cursor_shape = 2
 button_pressed = true
 text = "On"
 
+[node name="LowProcessorUsageLabel" type="Label" parent="HSplitContainer/VBoxContainer/ScrollContainer/RightSide/Performance/PerformanceContainer"]
+layout_mode = 2
+tooltip_text = "If this is toggled on, the application will slow down when not used. This helps lower CPU usage when idle."
+mouse_filter = 0
+text = "Lower processing speed when not in use"
```
OverloadedOrama, line 1065: This isn't entirely accurate. According to low processor mode's description in Godot's project settings, low processor mode doesn't redraw the screen if nothing changes visually. I recommend changing the text to just "Low processor mode" and the tooltip similar to Godot's tooltip, like "If this is enabled, the screen is not redrawn if nothing changes visually. This helps lower CPU usage and improve battery consumption."

Kiisu-Master: It only slows down redrawing, it doesn't actually prevent it.

Kiisu-Master: Could it be called "Slow on idle"? I think that would make more sense.

Kiisu-Master: We can also flip the setting and call it "Continuous redraw" similar to the godot editor setting

OverloadedOrama: I fear "slow on idle" could be confused with the "Pause application when it loses focus". I think having it be the same name as Godot is out safest bet

Kiisu-Master: In the Godot editor the setting is exposed as "Update Continuously" so i will change it to that.

OverloadedOrama: Sounds good to me!

OverloadedOrama (approved):
Looks good to me, thank you!

PR #1057, opened 2024-07-30
Make clipping masks work with Move tool and offset image dialog

Variable-ind:
https://github.com/user-attachments/assets/8761cb47-84d0-4dba-b313-57a1fb457a34


OverloadedOrama (approved):
Ideally we should find a way to not replicate the `blend_layers` method since a very similar method already exists in DrawingAlgos.gd, so perhaps we could re-write that somehow, but we could look into this for the future. Thank you!

PR #1069, opened 2024-08-09
Fix isometric grid calculations using ints instead of floats

kleonc:
Fixes a regression from #900 which changed some floating point calculations into integers.

Example broken setup|Before<br>(v1.0.1-stable)|After<br>(this PR)
-|-|-
L7FHB09F6s|Pixelorama_X81ZjEYRnx|Godot_v4 2 2-stable_win64_aPS7RxGvq3


OverloadedOrama (approved):
Thank you!

PR #1073, opened 2024-08-10
Update es_ES.po Error de traducción En Selecciones

devmark404:
Error de traducción https://github.com/Orama-Interactive/Pixelorama/issues/1072


OverloadedOrama:
Hello, thank you for the contribution! Translations should only change in [Crowdin](https://crowdin.com/project/pixelorama) and not directly in the files. I have already updated the translations there to fix #1072, and the fix will be applied once #1068 is merged.

PR #1075, opened 2024-08-11
Fix cursor blinking at the edge of canvas

Hamster5295:
## What's Wrong?
The cursor would blink at the edge of the painting canvas node at v1.0.1, switching between the state of CROSS and ARROW.
  
  
## What's the Cause?
The bug is caused by the additional `InputEventMouseMotion` triggered when calling `Input.set_default_cursor_shape(CursorShape)` in the function `_on_ViewportContainer_mouse_entered()` and `_on_ViewportContainer_mouse_exited()` of the script `res://src/UI/ViewportContainer.gd`, which leads to both of the functions being called infinitely, changing the cursor shape here and there.
  
  
## What does this PR do?
This PR substitudes the signals connected to `_on_ViewportContainer_mouse_entered()` and `_on_ViewportContainer_mouse_exited()` for the `_input(event)` callback, in a bid to handle the extra event manually, which fixs the situation mentioned above.
  
  
## Is it a Stable Bugfix?
The bugfix has been tested on the `DockableContainer/Main Canvas/ViewportandVerticalRuler/SubViewportContainer` Node, which is the canvas for painting, and seems to work **Fine**.   
However, since I'm not clear about the usage of `DockableContainer/Second Canvas`, the node which has also attached the script, so ***Further Testing is Required***

### DIFF
--- a/src/UI/ViewportContainer.gd
+++ b/src/UI/ViewportContainer.gd
@@ -2,9 +2,9 @@ extends SubViewportContainer
 
 @export var camera_path: NodePath
 
-@onready var camera := get_node(camera_path) as CanvasCamera
+var _mouse_inside = false
 
-var mouse_inside = false
+@onready var camera := get_node(camera_path) as CanvasCamera
 
 
 func _ready() -> void:
@@ -36,16 +36,16 @@ func _ready() -> void:
 func _input(event):
 	if event is InputEventMouseMotion:
 		if get_global_rect().has_point(event.position):
-			if !mouse_inside:
-				mouse_inside = true
+			if !_mouse_inside:
+				_mouse_inside = true
 				camera.set_process_input(true)
 				Global.control.left_cursor.visible = Global.show_left_tool_icon
 				Global.control.right_cursor.visible = Global.show_right_tool_icon
 				if Global.cross_cursor:
 					Input.set_default_cursor_shape(Input.CURSOR_CROSS)
 		else:
-			if mouse_inside:
-				mouse_inside = false
+			if _mouse_inside:
+				_mouse_inside = false
 				camera.drag = false
 				Global.control.left_cursor.visible = false
 				Global.control.right_cursor.visible = false
--- a/src/UI/ViewportContainer.gd
+++ b/src/UI/ViewportContainer.gd
@@ -12,27 +12,6 @@ func _ready() -> void:
 	material.blend_mode = CanvasItemMaterial.BLEND_MODE_PREMULT_ALPHA
 
 
-#func _on_ViewportContainer_mouse_entered() -> void:
-#return
-#camera.set_process_input(true)
-#Global.control.left_cursor.visible = Global.show_left_tool_icon
-#Global.control.right_cursor.visible = Global.show_right_tool_icon
-#if Global.cross_cursor:
-#Input.set_default_cursor_shape(Input.CURSOR_CROSS)
-#print("set cross")
-#
-#
-#func _on_ViewportContainer_mouse_exited() -> void:
-#return
-#print("exit")
-#camera.set_process_input(false)
-#camera.drag = false
-#Global.control.left_cursor.visible = false
-#Global.control.right_cursor.visible = false
-#Input.set_default_cursor_shape(Input.CURSOR_ARROW)
-#print("set arrow")
-
-
 func _input(event):
 	if event is InputEventMouseMotion:
 		if get_global_rect().has_point(event.position):
--- a/src/UI/ViewportContainer.gd
+++ b/src/UI/ViewportContainer.gd
@@ -5,22 +5,47 @@ extends SubViewportContainer
 @onready var camera := get_node(camera_path) as CanvasCamera
 
 
+var mouse_inside = false
+
 func _ready() -> void:
 	material = CanvasItemMaterial.new()
 	material.blend_mode = CanvasItemMaterial.BLEND_MODE_PREMULT_ALPHA
 
+#func _on_ViewportContainer_mouse_entered() -> void:
+	#return
+	#camera.set_process_input(true)
+	#Global.control.left_cursor.visible = Global.show_left_tool_icon
+	#Global.control.right_cursor.visible = Global.show_right_tool_icon
+	#if Global.cross_cursor:
+		#Input.set_default_cursor_shape(Input.CURSOR_CROSS)
+		#print("set cross")
+#
+#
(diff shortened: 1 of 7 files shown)


src/UI/ViewportContainer.gd
```diff
@@ -10,17 +12,41 @@ func _ready() -> void:
...
-		Input.set_default_cursor_shape(Input.CURSOR_CROSS)
-
-
-func _on_ViewportContainer_mouse_exited() -> void:
-	camera.set_process_input(false)
-	camera.drag = false
-	Global.control.left_cursor.visible = false
-	Global.control.right_cursor.visible = false
-	Input.set_default_cursor_shape(Input.CURSOR_ARROW)
+#func _on_ViewportContainer_mouse_entered() -> void:
```
OverloadedOrama, line 15: Please delete old code entirely instead of commenting it out 😅. Once that's done it should be good to get merged

Hamster5295: Sorry for that! I was considering about the circumstances where serious bugs are introduced and reverting to the original code is required, as it was not comprehensively tested at that moment 😹    
I'm going to remove those comments now 😸

OverloadedOrama: It's okay! If bugs are found and we decide to revert to the old code, we can always do that by looking at the git history of the file

OverloadedOrama (approved):
Looks good to me. To be honest I can't replicate the bug, but I don't notice any bug in the new code, so it should be good to be merged, since it fixes the issue on your end. Thank you!

OverloadedOrama:
Unfortunately I had to revert this change because it caused a bug where you couldn't rename layers anymore. Upon further testing the camera's process input is never set to `false` again, as it should be when the mouse is outside of the canvas, but even when I added a line that does that, it still seemed to cause weird behavior. We need to find another fix for the issue.

Hamster5295:
> To be honest I can't replicate the bug

That's kinda strange. Note that the blinking might appear to be the cross and the tool icons "getting dimmer" when the cursor is at the edge of the canvas due to visual persistence.  
In my case it behaves like this:  
  

I'll try to figure out whether it's caused by platform-specific issues.


> Unfortunately I had to revert this change because it caused a bug where you couldn't rename layers anymore. 

Get it. Sorry for the inconvenience! 😥

Variable-ind:
> > To be honest I can't replicate the bug
> 
> That's kinda strange. Note that the blinking might appear to be the cross and the tool icons "getting dimmer" when the cursor is at the edge of the canvas due to visual persistence. In my case it behaves like this:  [  ](https://private-user-images.githubusercontent.com/37259613/358814639-a5f2d557-2ac4-4902-b130-06eb0b1dcde5.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjM4NzU3NTYsIm5iZiI6MTcyMzg3NTQ1NiwicGF0aCI6Ii8zNzI1OTYxMy8zNTg4MTQ2MzktYTVmMmQ1NTctMmFjNC00OTAyLWIxMzAtMDZlYjBiMWRjZGU1LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA4MTclMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwODE3VDA2MTczNlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTA3M2MyMjQ3ODJlYmFlNzY0MjFjYTQ1OTk4YWM5YjhiMjg2ZWNhM2NhNGI3ZmZkNDJiY2MyNWMwMzRjZGEyYTgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.ezocBu2A1LIcRGShWMLYP44B3BLBJKnAUTQGLOOXIuQ) [ ](https://private-user-images.githubusercontent.com/37259613/358814639-a5f2d557-2ac4-4902-b130-06eb0b1dcde5.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjM4NzU3NTYsIm5iZiI6MTcyMzg3NTQ1NiwicGF0aCI6Ii8zNzI1OTYxMy8zNTg4MTQ2MzktYTVmMmQ1NTctMmFjNC00OTAyLWIxMzAtMDZlYjBiMWRjZGU1LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA4MTclMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwODE3VDA2MTczNlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTA3M2MyMjQ3ODJlYmFlNzY0MjFjYTQ1OTk4YWM5YjhiMjg2ZWNhM2NhNGI3ZmZkNDJiY2MyNWMwMzRjZGEyYTgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.ezocBu2A1LIcRGShWMLYP44B3BLBJKnAUTQGLOOXIuQ)
> 
> I'll try to figure out whether it's caused by platform-specific issues.
> 
> > Unfortunately I had to revert this change because it caused a bug where you couldn't rename layers anymore.
> 
> Get it. Sorry for the inconvenience! 😥

Hy I've noticed this bug on Linux as well, it's especially apparent when the ui scale is not 1

PR #1077, opened 2024-08-11
Implement support for group layer blending

OverloadedOrama:
Layers that belong to the same group are now being blended together into a single texture, which represents the entire group. Layers outside the group are not affecting the group's blended texture. This has several advantages:
- Group layers can have their own blend modes.
- Group layers can both act as a clipping mask and be used by a clipping mask.
- Group layers and cels now have their own opacity.
- Group layer effects are no longer being applied recursively to each of its children, but instead they are being applied once in the final blended texture. This allows us to use some modern non-destructive techniques, such as HD Index Painting with dithering support. https://www.youtube.com/watch?v=7Q36EyvaYG8
- A new "pass through" blending mode has been added that is only available for group layers. If that is selected, the children of the group are not being blended together, and instead it acts as if the group doesn't exist.

image
Recreating the HD Index Painting technique shown in the video I linked above. "Layer 1" is the black and white layer where we draw, "Layer 3" is the dithering pattern, and they are both being blended together by "Group 2", which contains a posterize and a gradient map layer effect.

https://github.com/user-attachments/assets/5c6324ea-221e-4d3e-ace7-6c1bfc5adb49

Group 3, containing a pixel and a 3D layer, is being used by Layer 2 which acts as a clipping mask.

Bugs fixed:
- Layer effects on group layers now automatically update the canvas the moment they are enabled/disabled.
- Changing a layer's blend mode, clipping mask status, opacity and a cel's opacity now automatically update the canvas, if these layers or cels are not currently selected.

This PR effectively makes group layers actually affect the blending process, rather than having group layers simply be for organization purposes and do nothing. ~~However, it might be a good idea to introduce a "pass through" blend mode that exists only for group layers, which would restore the previous behavior and make group layers have no effect on the blending process.~~

EDIT: Done

### DIFF
--- a/Translations/Translations.pot
+++ b/Translations/Translations.pot
@@ -2143,6 +2143,10 @@ msgstr ""
 msgid "Blend mode:"
 msgstr ""
 
+#. Found in the layer's section of the timeline, as a blend mode option, only available for group layers. If enabled, group blending is disabled and the group simply acts as a way to organize layers instead of affecting blending.
+msgid "Pass through"
+msgstr ""
+
 #. Adjective, refers to something usual/regular, such as the normal blend mode.
 msgid "Normal"
 msgstr ""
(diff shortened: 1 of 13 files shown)


OverloadedOrama:
Should be good to get merged, I didn't find any issues on my testing. Performance could use some work in the future though.

Issue #958, opened 2023-12-11
version 0.11.3 reset custom brush when switch another tool


alexTitakoff:
**Pixelorama version:**

0.11.3

**OS/device including version:**

mac os 10.14.6

**Issue description:**

Good afternoon!
Thanks to the developers for such a wonderful product. 
I found a bug. There was no such thing in version 0.11.2. The custom brush is reset when switching to another tool, in particular the eraser. It's a common movement to switch between an eraser and a brush.

This applies only to custom brushes

**Steps to reproduce:**

alexTitakoff:
Maybe it's related to this fix https://github.com/Orama-Interactive/Pixelorama/pull/916

PR #1078, opened 2024-08-13
Fix Orama-Interactive#958 (reset custom brush when switch another tool)

alikin12:
This should fix issue #958 (reset custom brush when switch another tool) by removing the childs safely from BrushPopup when project data is cleared.

### DIFF
--- a/src/UI/Buttons/BrushesPopup.gd
+++ b/src/UI/Buttons/BrushesPopup.gd
@@ -96,6 +96,7 @@ static func clear_project_brush() -> void:
 		"Background/Brushes/Categories/ProjectBrushContainer"
 	)
 	for child in container.get_children():
+		container.remove_child(child)
 		child.queue_free()
 		Global.brushes_popup.brush_removed.emit(child.brush)
 


OverloadedOrama:
Hello, thank you for the contribution! I'm not sure I understand what this change accomplishes, or how it's related to issue #958. I am trying to switch between projects with and without your change and I do not notice anything different. Can you explain how this fixes an issue?

alikin12:
Hello. First of all I want to thank you for your great work and the effort you and all other people who are involved in this project put into pixelorama. It's a very nice tool.

Regarding the issue and my fix I try to explain how to reproduce the issue and how the change fixes it:

Reproducing issue:
1. Create new brush
2. Save the project
3. Open the project a second time (or close the project and reopen it. Important is that the project is reloaded)
4. In the reloaded project: Select pencil tool and set brush to the new created brush. The brush should now be correctly selected
5. Change to any other tool
6. Change back to pencil tool. Now the self created brush is no more selected but the default Pixel brush

All this happens only when project was saved and reloaded. It does not happen without the reloading step.

Fix:
The issue appears because when a project is loaded, the brushes are added in the open_pxo_file method by add_project_brush. After that the change_project method is called which should clear all brushed and repopulate it again with the project ones. Without the fix the old brushes are not cleared directly (I think because queue_free() takes effect only on the next frame) and when the new brushes are added, the index which is set by get_index() is set to the wrong value as the old brushes still exist and the index has now an offset of the old brushs count. When switching the tool from another tool back to pencil, the brush with the wrong (not existing) index is tried to selected. As the child with this index does not exist, it fails and the default brush is selected.

Hope that helps :)

OverloadedOrama (approved):
Oh okay that makes perfect sense! Thank you very much for the detailed explanation and the solution to the problem!

PR #1079, opened 2024-08-13
Change german translation for "Unlink Cels"

alikin12:
Changed the german translation for "Unlink Cels" as the old translation means "Remove cel" which I interpreted as "cel/cel content will be deleted"


Variable-ind:
Hy, translations should be changed through crowdin

https://crowdin.com/project/pixelorama

PR #1085, opened 2024-08-23
Add Global layer buttons

alikin12:
Add global layer buttons to change visibility, lock or expand for all layers on the first level.
New added buttons in red rectangle:
grafik


OverloadedOrama (approved):
Nice addition, thank you!

PR #1093, opened 2024-08-29
Index Map

Variable-ind:
Implements index map layer effect, allowing you to use techniques such as this: https://www.youtube.com/watch?v=HsOKwUwL1bE

## (A basic example:)

https://github.com/user-attachments/assets/463b0ffe-0d4e-41c0-85aa-38bc24459d34

## Files used in above example
[Test Images.zip](https://github.com/user-attachments/files/16797998/Test.Images.zip)


OverloadedOrama (approved):
This is very cool! In the future we could implement using a different cel (from the same or another open project) as the map texture, so it would get updated automatically whenever the user makes a change, instead of having to save it as a png and load it manually, but for now this is fine. Thank you!

PR #1095, opened 2024-08-31
Set timeline layer and frame header to be fixed on top

alikin12:
This change fixes layer and frame header in the timeline window on top, so when you scroll down, you can still see the header.

I had to change the container structure to archive that. Also added some seperation between cell container and layer/frame container as I think that looks more clearly when scrolling. However don't know if it fits to what you have in mind.

Also did some other small changes which were needed for this new structure to get a better look (e.g. layer scroll bar is now always visible) and also fixed overlapping cel when selected on bottom.

grafik

Tested it under windows 10. Possibly someone can also check if the changes still look good on other operation systems.


OverloadedOrama (approved):
Very welcome changes! I tested them on Linux and they seem to work as intended, so this should be good to get merged. Thank you very much!

PR #1096, opened 2024-09-01
Added a way to modify shader textures

Variable-ind:
This will open the texture as a new project, saving that project won't save it locally but instead update the texture.

https://github.com/user-attachments/assets/e2e0b128-b12a-4f1e-b551-6d20fe46d668

[Files.zip](https://github.com/user-attachments/files/16827528/Files.zip)

### DIFF
--- a/src/Autoload/Global.gd
+++ b/src/Autoload/Global.gd
@@ -704,17 +704,6 @@ func _init() -> void:
 	)
 
 
-func modify_texture_resource(image: Image, resource_name: StringName, update_callable: Callable):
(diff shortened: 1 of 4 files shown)


src/Autoload/Global.gd
```diff
@@ -693,6 +704,17 @@ func _init() -> void:
 	)
 
 
+func modify_texture_resource(image: Image, resource_name: StringName, update_callable: Callable):
```
OverloadedOrama, line 707: Maybe this method should be in ShaderLoader instead, since it's only called from there

OverloadedOrama (approved):
Awesome, thank you!

PR #1100, opened 2024-09-04
Performance boost for spritesheet export

Variable-ind:
1. Added some static typing (i heard cases like `: bool`, `: Vector2i` etc works faster than `:=` so i changed them)
2. Added a recycling system for spritesheets, (old spritesheets get reused and only the different parts get changed). this makes changes to nearby values (e.g slider moved from `rows = 23` to `rows =  24`) almost instantaneous.
3. Added a slight delay between Valueslider (Yeah, i changed row/column `SpinBox` to a `Valueslider` :grin: ), this delay works the same as the Live Preview option in some ImageEffects, making it more smooth.

### DIFF
--- a/src/Autoload/DrawingAlgos.gd
+++ b/src/Autoload/DrawingAlgos.gd
@@ -25,39 +25,39 @@ func blend_layers(
 	only_selected_cels := false,
 	only_selected_layers := false,
 ) -> void:
-	var frame_index := project.frames.find(frame)
+	var frame_index: int = project.frames.find(frame)
 	var previous_ordered_layers: Array[int] = project.ordered_layers
 	project.order_layers(frame_index)
 	var textures: Array[Image] = []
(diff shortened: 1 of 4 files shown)


OverloadedOrama:
> Added some static typing (i heard cases like `: bool`, `: Vector2i` etc works faster than `:=` so i changed them)

I don't think this is true, where did you hear this? Both `: bool` etc and `:=` should have the exact same performance, since they are both examples of static typing. In most cases, I personally prefer inferred static typing. I think specifying the variable type explicitly often creates unnecessary duplications and doesn't really make the code any more readable.
For example, `var image: Image = Image.new()` or `var some_bool: bool = false` don't really need explicit type declarations, since it's already obvious that they are an image and a boolean respectively, so I would avoid specifying their type and just use `:=` instead.

Variable-ind:
Closing this in favor of a0c7dd452761a077ff2f07cd4489248c3ae4e752
The ValueSlider will be added in the future.

PR #1104, opened 2024-09-14
 Set dynamics dialog to correct size after something is made visible …

Variable-ind:
…or hidden


src/UI/GlobalToolOptions/DynamicsPanel.gd
```diff
@@ -29,6 +29,13 @@ func _ready() -> void:
 	size_velocity_button.toggled.connect(
 		_on_dynamics_toggled.bind(size_velocity_button, SIZE, Tools.Dynamics.VELOCITY)
 	)
+	for child: Control in $VBoxContainer.get_children():
+		## Resets the y-size to an appropriate value
+		child.visibility_changed.connect(func ():
```
OverloadedOrama, line 34: Not sure why the static checks are failing, but perhaps it would be best to make this a proper method rather than a lambda since it has many lines.

Variable-ind: Done :+1:

OverloadedOrama (approved):
Thanks!

PR #1105, opened 2024-09-14
improved rotate/flip brush UI

Variable-ind:
After


OverloadedOrama:
I think it would look better if the `:` was in the same label as the text and not its own label. It would also be consistent with the rest of the UI, in places such as other tool options like the 3D Shape Edit tool and image effects like gradient.

Variable-ind:
New look After Changes:

Screenshot from 2024-10-14 16-01-02

https://github.com/user-attachments/assets/dbd92e05-eb50-45a5-830a-93d6c2d8ae47

OverloadedOrama (approved):
Awesome, thank you!

PR #1107, opened 2024-09-14
ColorReplace Shading Mode

Variable-ind:
This pr adds a color replace shading mode similar to aseprite

## Demonstration of Aseprite's shading tool:

https://www.youtube.com/shorts/ty3qI4v5M0Y

## Demonstration of Pixelorama's shading tool

https://github.com/user-attachments/assets/d6de1136-dd51-4c9b-a8e2-2afee5925a05

### DIFF
--- a/src/Tools/DesignTools/Shading.gd
+++ b/src/Tools/DesignTools/Shading.gd
@@ -321,10 +321,12 @@ func update_brush() -> void:
 	$ColorInterpolation.visible = false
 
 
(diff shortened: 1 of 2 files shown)


OverloadedOrama (approved):
Looks good! In the future if we implement the ability to select multiple color swatches, it may be a good idea to replace the "colors right" slider and instead use the selected swatches, but for now this should be good enough. Thank you!

PR #1108, opened 2024-09-15
Fixed a colorpicker bug (fixed inconsistent raw values)

Variable-ind:
I noticed that the raw values slightly change when you select color through color picker, draw, then pick the color.
This pr fixes this.

https://github.com/user-attachments/assets/7ad6cc40-9f18-452d-bf74-eb852551d947


OverloadedOrama (approved):
Thanks!

PR #1113, opened 2024-09-27
Allow clipping to selection during export

Variable-ind:


src/Autoload/Export.gd
```diff
@@ -288,6 +289,14 @@ func process_animation(project := Global.current_project) -> void:
 		else:
 			var image := Image.create(project.size.x, project.size.y, false, Image.FORMAT_RGBA8)
 			image.copy_from(blended_frames[frame])
+			if erase_unselected_area and project.has_selection:
+				var clipper = ShaderImageEffect.new()
```
OverloadedOrama, line 293: Do we need a shader for this? Wouldn't `Image.blit_rect_mask()` also work?

Variable-ind: changed

OverloadedOrama (approved):
Awesome, thank you!

PR #1116, opened 2024-10-11
somewhat fix transparency for floating window

Variable-ind:
Transparency works as expected when Main canvas is part of MAIN window.
but if it gets detached then it behaves un-expectedly in some ways.
- Transparency shader still remains in the Main window (so i temporarily remove it when main canvas is a floating window)
- The transparency didn't work for Floating window (It works now, but only the transparent checker gets modulated, the rest of window is fully transparent which was kinda un-avoidable).

I don't see a reason why a user would decide to detach the Main canvas (Plus enable transparency) but if they do then i hope this will minimize the annoyance


OverloadedOrama (approved):
Thanks!

PR #1117, opened 2024-10-12
Add alpha erase blend mode

Variable-ind:
Works same as the Erase mode in https://docs.gimp.org/en/gimp-concepts-layer-modes.html (`Figure 8.12. Example for layer mode “Erase”`)

An "Erase" layer blend mode or effect would easily skip a step of manual labor when, for example, making opaque shapes with transparent icons within.

https://github.com/user-attachments/assets/559387a8-4133-47b9-8e1b-95c01601009c

https://github.com/user-attachments/assets/2817253f-fdb3-4a88-ab59-797885acf717


OverloadedOrama:
Looks good! If we want to be consistent with other software though, I think we should keep the name "Subtract" as it is, and the new mode should be called "Erase", as they are not quite the same.  Erase should also be together with Normal, as its not considered an Inversion type of blend mode. In the documentation you linked it is listed under the Normal layer modes group, while Subtract is listed under the Inversion layer modes https://docs.gimp.org/en/layer-mode-group-inversion.html

Variable-ind:
Done

OverloadedOrama (approved):
Looks great, thank you!

PR #1118, opened 2024-10-13
Update RegionUnpacker.gd

HuanWuCode:
Fix typo. for element: int in side_b: This line should be for element in range(side_b): since side_b is likely an integer.


Variable-ind:
Hy 👋, this isn't a typo
Godot can loop over integers so a range() isn't required

PR #1120, opened 2024-10-14
Add a preference to share options between tools

Variable-ind:
Supersedes #989 
- Adds a Global tool option that enables sharing of common configs between the two tools. Sync is done whenever `save_config()` is called.

(Other necessary changes)
- Flip/rotate options now get saved between sessions
- Fill of the curve tool now gets saved between sessions


OverloadedOrama (approved):
After some discussion, we thought it would be best to make this an option in the preferences instead of a button in the global tool options, as this is an option that will most likely not be toggled on and off very often. It's best to keep such options in the preferences to avoid cluttering the UI too much.

Thank you!

PR #1122, opened 2024-10-15
Allow multiple Grids

Variable-ind:
- The next grid is drawn over the previous grid so first draw a smaller grid, then a bigger grid and so on
   -  Snapping is done by the grid at layer zero (In the future we may allow snapping to all grids but for now, bear with me)
   - Maximum 10 layers are allowed
    
Testing is required

https://github.com/user-attachments/assets/2f57e4a7-6425-4b74-904e-8bee7b098cf7


OverloadedOrama (approved):
Looks good to me. Did some testing and couldn't find any issues. Thank you!

PR #1131, opened 2024-11-08
Add a way to get autoloads through the api

Variable-ind:
This allows accessing autoloads by placing an `autoload()` method in their related apis
(for the ImportApi, there are actually 2 available autoloads so it slightly deviates from the above naming convention)

this won't cover (as i don't think these are not important, and we don't have a category for it)
- Html5FileExchange
- Keychain


OverloadedOrama (approved):
Thanks!

PR #1134, opened 2024-11-18
Add a text tool

OverloadedOrama:
Adds a new text tool. Partially implements #589. The reason I delayed adding a text tool so much was because the plan was to first implement vector layers, and then add the text tool for non-destructive text editing. But this is taking longer than expected, and Pixelorama needs a way to add text, so I made a destructive text tool. This means that once you confirm the text, you cannot edit it later.

https://github.com/user-attachments/assets/8cee2c6f-a60d-43a6-8b1e-03b5ceda8536

Usability-wise, it acts similarly to Photoshop/Photopea, where you click somewhere on the canvas and the text will appear there, and dragging anywhere outside the text area will move the text around. To insert text, simply click inside the text area and start typing. To confirm or cancel the changes, you have to click on the confirm/cancel buttons in the text tool options.


OverloadedOrama:
It could probably use some more work and feedback, but for now I think it's good enough to get merged and we can make changes later.

PR #1138, opened 2024-11-21
Optimize the resize of the selection when the selection is an ellipse.

NIyueeE:
If the selection is an ellipse, resize it as a standard ellipse.


NIyueeE:
https://github.com/user-attachments/assets/2cb1163a-243d-4918-9b82-8e8cddb89b12

NIyueeE:
https://github.com/user-attachments/assets/10f508f0-aeca-4cc2-a3b3-109f1c888eb9

OverloadedOrama:
Hello, thanks for the contribution! Can you explain why this is an optimization? From the look of it, a lot more code is added to `resize_bitmap_values()` instead of just having `smaller_image.resize(new_size.x, new_size.y, Image.INTERPOLATE_NEAREST)`. And it always seems to call methods such as `DrawingAlgos.get_ellipse_points()`, which aren't exactly the fastest, especially with their current GDScript implementation.

Moreover, this changes the behavior of the resizing of elliptical selections, which results in the selection no longer matching the content it is resizing, as shown in the video:

https://github.com/user-attachments/assets/754cc744-963f-499e-8051-dcc187b97252

NIyueeE:
> Hello, thanks for the contribution! Can you explain why this is an optimization? From the look of it, a lot more code is added to `resize_bitmap_values()` instead of just having `smaller_image.resize(new_size.x, new_size.y, Image.INTERPOLATE_NEAREST)`. And it always seems to call methods such as `DrawingAlgos.get_ellipse_points()`, which aren't exactly the fastest, especially with their current GDScript implementation.
> 
> Moreover, this changes the behavior of the resizing of elliptical selections, which results in the selection no longer matching the content it is resizing, as shown in the video:
> 
>  2024-11-24.14-17-06.mp4
I misrepresented myself.
I just want to add the feature to customize the length and width of the ellipse. So I came up with what I thought would work.
Actually, I'm not familiar with GDSript and the whole project, so there's probably a better way to implement this.
Hope the feature will be added future🙂( or just maybe worked with other ways?), the pull request will be closed.

PR #1139, opened 2024-11-23
fixed recorder label not updating when project is changed

Variable-ind:
the recorder label wasn't updating when:
- we changed project (with both projects recording)
- when the recording was just started (the text of label is meant to be empty at that point)

also changed the recorder icon (to a slightly higher resolution one), the old icon was somehow not being displayed correctly (see image below)

### Old:
Screenshot from 2024-11-23 16-35-44

### New:
Screenshot from 2024-11-23 16-35-08


OverloadedOrama (approved):
Thanks!

PR #1142, opened 2024-11-25
Grid patch

Variable-ind:
- Fixed second grid not appearing to be removed when first grid has default values. this was because queue_redraw was not called due to the app thinking nothing had changed.
- fixed reset button of color picker not set correctly when we switch to another grid
- Now, whenever you create a new grid, the new grid will be twice the size of previous grid (and with a different color). this is done because i felt that having all grids being being of same size and color may cause confusion for users when they try to change color of a grid not seeing it's changing (due to being covered by grids above it).

### DIFF
--- a/src/Autoload/Global.gd
+++ b/src/Autoload/Global.gd
@@ -752,7 +752,11 @@ func _ready() -> void:
 
 
 func update_grids(grids_data: Dictionary):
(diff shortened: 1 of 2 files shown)


OverloadedOrama (approved):
Thanks!

PR #1143, opened 2024-11-26
Added a way to see index of color (plus bugfixes)

Variable-ind:
### Added:
An index preview (same as what was used in voxelorama) that displays after the same zoom as pixel grid
NOTE: i attempted to make them visible at all zoom levels but it caused a visual artifact at smaller zoom levels (see video below) so i thought it was best to restrict it up to a given zoom level.

https://github.com/user-attachments/assets/d5bd838f-c25a-4464-a01d-60921b079eb6

### Fixes:
- [Index mode] If the color we are drawing is already selected in the palette, then it's index now takes priority (this solves any issues with palettes having multiple instances of the same color)
- [Palettes] Fixed more than one swatch selected in grid if the same color available in an earlier swatch.
- [Index mode] Fixed wrong index drawn when moved to an empty swatch (when we drag and drop to an empty swatch, after updating palette, we were getting two entries of that color because the last position was not erased)
- [Index mode] When populating the palettes array, the color added was slightly wrong (i think the reason is same as https://github.com/Orama-Interactive/Pixelorama/pull/1108). due to this the `palette.has(color)` check in `set_pixelv_custom` was almost always false even if the color was there in the palette

### DIFF
--- a/project.godot
+++ b/project.godot
@@ -921,7 +921,7 @@ right_text_tool={
 "events": [Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":-1,"window_id":0,"alt_pressed":true,"shift_pressed":false,"ctrl_pressed":false,"meta_pressed":false,"pressed":false,"keycode":0,"physical_keycode":65,"key_label":0,"unicode":97,"location":0,"echo":false,"script":null)
 ]
 }
-show_color_indices={
+show_pixel_indices={
 "deadzone": 0.5,
 "events": []
 }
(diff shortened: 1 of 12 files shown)


OverloadedOrama:
Looks pretty good! I think it may be better to not show the indices at all for transparent pixels, since they are all 0. It would make the canvas less noisy, and the pixels with non-zero indices would be easier to read.

OverloadedOrama (approved):
Looks good, thank you!

PR #1144, opened 2024-11-29
When picking color, ensure that the correct index is selected in palette

Variable-ind:
other changes:
- If drawing over same color in index mode, check and update index as well
- revert an optimization i did in https://github.com/Orama-Interactive/Pixelorama/pull/1143 (commit: https://github.com/Orama-Interactive/Pixelorama/pull/1143/commits/cf32c8812a86c520d1580224f7ed0f2362d590cc)


OverloadedOrama (approved):
Thanks!

PR #1146, opened 2024-11-30
Implement tilemap layers

OverloadedOrama:
Implements a new type of layer, tilemap layers. Tilemap layers are a type of pixel layers that are divided by grid cells, where each cell is mapped to a tile in a tileset. Tilesets are stored in projects, and multiple tilemap layers can use the same tileset.

This is a huge new feature that I wanted to try to have ready for v1.1, and while I did do a lot of tests before opening this PR, more testing would be greatly appreciated. I plan on merging this PR as soon as possible to make it easier for more people to test it.

https://github.com/user-attachments/assets/3cde8e41-ed11-4e23-992b-00d55ff0ba74

Current features:
- Users can set the size of the tiles in the tileset while creating it. Tilesets are being created when adding a new tilemap layer, but tilemap layers can also re-use existing tilesets.
- Manual tile editing mode, which modifies individual tiles without re-ordering the tileset, and the modifications get immediately applied to other cells in the tilemap which are mapped to the same tile. Modifications to tiles also get applied to other tilemap layers and cels that have the same tileset.
- Auto tile editing mode, which automatically creates new tiles if they don't already exist, it modifies tiles that only exist once, and it deletes unused tiles.
- Stack tile editing mode, which always creates new tiles every time the user makes a modification, without affecting the previous tiles.
- Draw tiles mode, which directly places tiles from the tileset in the tilemap, instead of pixels. Can be used by pencil, eraser, bucket, shape and selection tools.
- Resizing selections resizes tile indices, when drawing tiles mode is enabled.
- Rotation and flipping support for when placing tiles in the tilemap.
- A new tileset panel which contains all of the above options, plus the tileset of the currently selected tilemap layer. If no tilemap layer is selected, this panel is invisible.
- The first grid always has the same size as the tile size, if a tilemap layer is selected, regardless of the size set in the Preferences. Other grids are not affected.
- Show the index of each tile in the tilemap layer when <kbd>Control</kbd> is held.
- Load images as tilesets.
- Duplicate and delete tilesets from the project properties dialog.
- Change which tileset the tilemap is using from its layer properties.
- Layer/frame cloning, cel linking, drawing on multiple cels, undo/redo, layer merging, indexed mode, pxo saving/loading, image/layer effects, resizing should all work.

https://github.com/user-attachments/assets/c95722d8-491e-42e9-91fb-07d8d46f0471

Stuff that can be added later:
- Use icons instead of text for the buttons in the tileset panel.
- Add a small icon in the layer button in the timeline to differentiate it from other layer types. Should also be done for group and 3D layers.
- Move tool on tilemap layers should move the tilemap offset.
- Support isometric tiles. Perhaps other shapes as well, such as hexagons.
- Export to Godot TileSets and TileMapLayers.


OverloadedOrama:
https://github.com/user-attachments/assets/f727e6a5-375a-472b-b96b-6363cf2eb52b

Managed to make selection tools and resizing work with the draw tiles mode. I think the tilemap layer system is pretty feature complete now, I'll merge this PR so Nightly version users can test it and give feedback.

PR #1150, opened 2024-12-14
Renamed the selection "tile mode" to "wrap strokes"

Variable-ind:
As it is a more accurate description of what it does


OverloadedOrama (approved):
Thanks!

PR #1167, opened 2025-01-24
Fix always_on_top for sub-dialogs

Variable-ind:
Fix Create Folder sub-dialog of Export > Browse being always behind (thus making it unusable).

This shall fix the problem on other locations as well. As long as they are included in the FileDialogs tag (in Main.tscn)


Variable-ind:
I noticed there are some other cases where the popups of option buttons get hidden behind the window if the window has always_on_top enabled. (like the `LayoutFrom` option button in ManageLayouts.tscn)

so i'm converting this to a draft and will modify this later to include those fixes as well

Variable-ind:
Okay, so i managed to fully fix it
(Some comments about the PR):
- FixInternalDialogs is a dedicated group which only contains items which need correction (once they are corrected they are removed from the group).
- FixInternalDialogs is a `Global` group instead of a `Scene` group.
   - This is to avoid the hassle of creating the same group again and again for each scene (Plus they can be change to scene groups later anyway, if the need arises).
   -  This also make it easier to undo this PR if it gets fixed in Godot (i checked in Godot 4.4 but it wasn't fixed yet)
- The code for correction is moved to the `Dialog` class because:
    - **(Short version):** It's a more general solution.
    - **(Long version)** Even if currently only `Export` and `Save` dialogs have this bug (i haven't seen it anyplace else, besides the now removed `ManageLayouts.tscn`), we may in the future encounter this in one of the Dialogs (like we did in the `LayoutFrom` OptionButton of `ManageLayouts.tscn`), And since Dialogs are not "immediately" part of the scene, we have to fix it's internal dialogs WHILE it's being added to tree.

Related stuff:
https://github.com/godotengine/godot/issues/95096
https://github.com/godotengine/godot/issues/92848

src/UI/TopMenuContainer/TopMenuContainer.gd
```diff
@@ -99,6 +99,21 @@ class Dialog:
 		node = scene.instantiate()
 		if is_instance_valid(node):
 			Global.control.get_node("Dialogs").add_child(node)
+			fix_internal_dialogs()
+
+	static func fix_internal_dialogs() -> void:
```
OverloadedOrama, line 104: I'm a bit confused as to why this is here and not in Main.gd, since this is only called once. The only windows in FixInternalDialogs are the ExportDialog, OpenSprite and SaveSprite, and once this method is called, they are being removed from the group, so calling this method on Dialog's `instantiate_scene` doesn't actually do anything since there are no nodesl eft in the FixInternalDialogs group. Did you mean to add new nodes to that group when they get instantiated?

Variable-ind: It's a more general solution
> since there are no nodes left in the FixInternalDialogs group.

Yes, they aren't any at present but this may not be the case in the future where dialogs themselves have internal dialogs (in the form of OptionButtons etc.. like i mentioned above).

In that case there WOULD be a non-zero number of nodes in the `FixInternalDialogs` group (as the Dialog **just** got added).

Variable-ind: (for the sake of record) After some discussion we agreed to move the code to the ExportDialog.gd

OverloadedOrama (approved):
Thanks!

PR #1172, opened 2025-02-04
Fix layouts overwriting the position info of pannels, added by extensions.

Variable-ind:
The problem was that the layout change was done too early, before the extension adding the panel.
As the panel always initially gets added to the left most side (which is different form it's last closed position, where we moved the panel), it gets counted as a change and the layout overwrote/reset the panel's position


OverloadedOrama (approved):
Looks good to me, thank you!

Issue #1153, opened 2024-12-16
[UX] There is no way to swap right and left tool with a key press. 


ludrol:
**Pixelorama version:**
1.0.5-stable

**OS/device including version:**
Windows 10 64x
Huion H640p graphics tablet

**Issue description:**
I have a tablet that activates left click on touch and right click as an additional button. 
When right clicking the pen is always hovering that lacks precision. It is impossible to use right tool with my graphics tablet. 

**Steps to reproduce:**

Have a graphics tablet that right clicks during hover. 

Can be emulated with tablet that tracks during hover and keybind the right tool to a keybord key. 

Erasing seems impossible if right tool is an eraser

OverloadedOrama:
Implemented by #1173.

PR #1173, opened 2025-02-05
Adds ability to swap tools (Shift + x as an initial shortcut)

Variable-ind:
Implements #1153


OverloadedOrama (approved):
Looks good to me, thanks!

c0d3r9:
Thanks for the implementation, but it works not 100% correct.
As example: press the eraser shortcut (i set it to H for the test) than the eraser is active as long as i hold H. After release H it swap back to the Pencil i.e.. 
Now the problem: If i wanna eraser something while i paint with the pencil next i press H and it happens nothing.
So the Quick shortcut seem not to work while working active with another tool. To activate the quick tool you have to release the left mousebotton or hold the pen away from tablet.
As a tip, the "swap tool" for the right and left mousebutton works like a charm.
This behaviour is needed for the new quick tools feature.
Thanks alot.

Variable-ind:
> This behaviour is needed for the new quick tools feature.

Hy 👋, in this case I don't think this PR is the problem, as this in not the one that implements Quick tools. Also it'll be more proper to open a separate issue for the problem instead 😁

PR #1177, opened 2025-02-14
Revised Extension crash failsafe

Variable-ind:
The Main philosophy behind the failsafe is to make sure PIXELORAMA loads up without an issue:

- This revision adds an additional algorithm that check if extensions crash DURING the time they are enabled. This will make it almost impossible for an extension to keep crashing pixelorama (it will do the necessary actions to auto recover in the next session)

~~(This isn't playing well if multiple instances are running, till i figure it out it's a draft)~~

- If another instance is run, How does it decide if the previous session crashed?
    - The system uses a `sane_timer` that pings the monitor file every 10 seconds. If the **Monitor.ini file exists** AND the time difference between the **last updated time** and the time of creation of a new instance is **greater than the `wait_time` (10 seconds)**, it would mean that for some reason, the the previous instance failed to update the time (which would be interpreted as s crash).

I would have liked to make it work without the timer but this is the ONLY way this can work

### DIFF
--- a/src/HandleExtensions.gd
+++ b/src/HandleExtensions.gd
@@ -55,9 +55,7 @@ func _ready() -> void:
 	prev_damaged_extensions = remove_damaged()
 	if !prev_damaged_extensions.is_empty():
 		if prev_damaged_extensions.size() == 1:
-			var error_text = (
-				"A Faulty extension was found in previous session:\n%s\nIt will be moved to:\n%s"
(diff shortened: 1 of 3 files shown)


Variable-ind:
This pr is now ready for review

OverloadedOrama (approved):
Looks good to me, thank you!

PR #1189, opened 2025-02-28
Fix Multi Screen Bug

### LINKED ISSUE #1188: Multi screen: The program will crash on splash screen outside of the current screen if the previously cached screen aren't available
**Pixelorama version:** 1.0.5-stable

**Where did you download Pixelorama from?** GitHub Releases

**OS/device including version:** OS: Windows 11 Pro 10.0.26100 Build 26100, Graphics: Intel(R) Iris(R) Xe Graphics

**Issue description:** 
Hey guys! I was using Pixelorama at work on my extra monitor (i have a laptop), once i finished using it, i closed it and later opened it again but without using this extra monitor but something happened (or didn't).

After the splash screen, i could not see the window and neither close it normally, needed to use the Task Manager.

Even if i focus on the program on task bar, i couldn't move the window around using `Alt + Arrows` like i'd normally do, it was annoying.

I can't test on my linux desktop at home because i don't have an external monitor, but you guys could and say if it happens there too.

I've already checked out the source code and made a small change that could fix it, at least i tried and it worked.

I'm opening a pull request along with this issue.

Thanks in advance, this tool is great!

**Steps to reproduce:**
1. Open the program
2. Move it to an external monitor
3. Close the program
4. Disconnect the external monitor
5. Try to open the program again and voila!

dev-joaovitor:
fixes #1188

### DIFF
--- a/src/Main.gd
+++ b/src/Main.gd
@@ -269,8 +269,14 @@ func _setup_application_window_size() -> void:
 
 	if OS.get_name() == "Web":
 		return
+
+
 	# Restore the window position/size if values are present in the configuration cache
 	if Global.config_cache.has_section_key("window", "screen"):
+		# Restore the window configuration if the screen in cache is not available
+		if (DisplayServer.get_screen_count()
+			< (Global.config_cache.get_value("window", "screen") + 1)):
+			return
 		get_window().current_screen = Global.config_cache.get_value("window", "screen")
 	if Global.config_cache.has_section_key("window", "maximized"):
 		get_window().mode = (
--- a/src/Main.gd
+++ b/src/Main.gd
@@ -270,12 +270,13 @@ func _setup_application_window_size() -> void:
 	if OS.get_name() == "Web":
 		return
 
-
 	# Restore the window position/size if values are present in the configuration cache
 	if Global.config_cache.has_section_key("window", "screen"):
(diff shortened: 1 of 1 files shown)


src/Main.gd
```diff
@@ -269,8 +269,14 @@ func _setup_application_window_size() -> void:
...
 	if OS.get_name() == "Web":
 		return
+
+
 	# Restore the window position/size if values are present in the configuration cache
 	if Global.config_cache.has_section_key("window", "screen"):
+		# Restore the window configuration if the screen in cache is not available
+		if (DisplayServer.get_screen_count()
+			< (Global.config_cache.get_value("window", "screen") + 1)):
+			return
```
OverloadedOrama, line 279: Not too sure about the return here, that will prevent the rest of the method code from executing. I think a better solution may be doing this:
```
var screen = Global.config_cache.get_value("window", "screen")
if DisplayServer.get_screen_count() >= screen + 1:
	get_window().current_screen = screen
```

dev-joaovitor: My first attempt was this one, but if the external monitor resolution is different than the current, the position may be a problem too.

In my case even if the screen was correct, the laptop resolution differ from the external one, that's why i've chosen to not apply any of the cache values

OverloadedOrama (approved):
Thank you!

PR #1194, opened 2025-03-09
[bugfix + Gizmo] Improve the look of 3d movement gizmos

Variable-ind:
- Fix a bug where axis sometimes got inverted when scaling (fixed by using `normalized()`)

https://github.com/user-attachments/assets/df462902-f7d4-4efc-acb8-932fa183c4bd

- The axis lines have different width depending on their position, this adds a little bit of perspective/depth to the gizmos
- The axis lines stay visible when rotation is active (this is more pleasing to the eye, plus the user can easily get an intuitive understanding of what's going on for symmetrical shapes like cylinder, sphere, etc)
- The X, Y, Z text was too large so `CHAR_SCALE` was decreased from 16 to 10 (This is not shown in the picture)

Screenshot from 2025-03-09 23-10-45


OverloadedOrama (approved):
Thanks!

PR #1208, opened 2025-03-29
DRAFT: Implement C++ Solutions for Algorithms

Schweini07:
This PR is a draft for a potential C++ implementation for heavy-compute algorithms and others in Pixelorama. Currently the PR is more so a proof of concept, as I would like to discuss first how to best integrate GDExtension into Pixelorama before continuing further.

Some things I am not really happy/unsure about:
- Right now the gdextension directory is placed in the root of the project which I think should be fine, but I am unsure if it might be better of somewhere else.
- To call the DrawingAlgosCpp functions I needed to create an Autoload wrapper through which  the C++ class is called. I am not really happy about this, but I don't think there is a better way.
- There is no Array[Vector2i] in godot-cpp (https://github.com/godotengine/godot-proposals/issues/5486) so I every time such a variable is used it has to be retyped, which is a real nuisance.

Let me know your thoughts about what might need some improvement, before I continue with this :slightly_smiling_face:


Schweini07:
> I think it would be better if it was somewhere else, maybe in the addons folder. In fact, the way I originally thought about it would be to have a completely new repository for the GDExtension, and Pixelorama would just have the compiled binaries (.so, .dll, .dylib etc). This would also allow other projects to also use these algorithms if they wanted to. I actually had started working on such a project some months ago, but I did not continue working on it since then. I just made the repository public, feel free to take a look and if you agree with this approach, you can contribute PRs there. https://github.com/Orama-Interactive/Godot-Atelier Otherwise if you think it would be better to have the C++ code in this repository and make it Pixelorama specific, we could just keep it here.

I think that is a great idea! That should void one of the concerns I had, that new contributors to Pixelorama would have to manually compile the C++ code themselves and therefore creating a barrier of entry. I assume the binaries can then also be build through Github Actions, right?

I've also begun working on your repository, I will hopefully be able to make a PR soon :).

> We could make DrawingAlgosCpp inherit from RefCounted instead of node, and have its functions be static. I think it would allow us to just call these functions in GDScript just by doing DrawingAlgosCPP.function_name(), but I haven't tested it yet.

Right, I didn't even think of that :D. I just tested it and it works without problems, that should make everything a lot more convenient.

> I think TypedArray<Vector2i> should work in C++.

Thanks! The lack of godot-cpp documentation made me just believe that there is no such type. Glad to be proven wrong.

OverloadedOrama:
> I assume the binaries can then also be build through Github Actions, right?

Yeah I believe so, perhaps we could use this: https://github.com/godotengine/godot-cpp-template/blob/main/.github/workflows/builds.yml

PR #1211, opened 2025-04-01
fix section not found (preferences) error

Variable-ind:
For first time downloads pixelorama would give error about section not found (because of nonexistent config.ini)

This will check if the section exists, before trying to get it's keys


OverloadedOrama (approved):
Thank you!

PR #1214, opened 2025-04-07
Added a way for users to report faulty extensions

Variable-ind:
Some times when an extension turns out to be faulty/outdated this will allow users to report it, so that they are disabled in repository and their Authors notified of updating them.


OverloadedOrama (approved):
Thank you!

PR #1216, opened 2025-04-11
Added a single bezier mode in curve tool

Variable-ind:
this mode will allow users to make single Bezier curves similar to Aseprite.


OverloadedOrama (approved):
Looks good to me, thank you!

PR #1222, opened 2025-04-20
Improve the preview of curve Single mode a bit

Variable-ind:
makes it more accurate to what it's actually doing while in the MIDDLE_A state.


OverloadedOrama (approved):
Thanks!

PR #1240, opened 2025-05-13
Make pixel grid redraw on visibility toggle

VernalUmbrella:
When the (non-pixel) grid is toggled, it immediately updates to show the change.
Before, when the pixel grid was toggled, it would remain visible/hidden until updated by zooming. Add a queue_redraw call to fix.


OverloadedOrama (approved):
Thank you!

PR #1242, opened 2025-05-17
Fix #1059 by disabling DragNDropPanel when "Moveable Panels" is disabled

### LINKED ISSUE #1059: Panels are moveable even if Moveable Panels option is turned off
**Pixelorama version:** master

**OS/device including version:** Arch Linux, KDE Plasma 6.2, Wayland

**Issue description:**

If the `Window > Moveable Panels` option is disabled it's still possible to move panels which share a place with different panels and thus their tab is visible. This can lead to an unwanted rearrangement of a current layout.

VernalUmbrella:
Closes #1059.

### DIFF
--- a/addons/dockable_container/dockable_container.gd
+++ b/addons/dockable_container/dockable_container.gd
@@ -97,18 +97,19 @@ func _ready() -> void:
 
 
 func _notification(what: int) -> void:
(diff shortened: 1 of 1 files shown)


OverloadedOrama (approved):
Thank you!

PR #1244, opened 2025-05-20
Added ability to double-click on preview to get at the same point on canvas

Variable-ind:
This removes the extra steps involved like zooming out the zooming back in.


piotrtrkostrzewski94:
Very useful, thank you very much again :)

OverloadedOrama (approved):
Thanks!

PR #1252, opened 2025-05-26
Use pixelated grid for isometric grid

Variable-ind:
This is more consistent and accurate than the old grid for pixelart.

https://github.com/user-attachments/assets/aec3ebb7-3acf-4db9-95e2-3506408a4bb7


OverloadedOrama (approved):
Thank you!

PR #1253, opened 2025-05-26
Refactor of tilemap cels

Variable-ind:
## Added:
- Added ability to edit individual tiles even in place_only mode.
- Allow support for batch removal of unused tiles.

## Fixed:
- The current implementation had some issues because it treats all tiles as equal size when in reality they are not.
- Improved tile generation system.
- Improved Eraser tool's tilemap behavior
- FIX selection tool's behavior.
     - Fixed behavior of isometric selection deletion was broken for isometric tilemap.
     - Fixed Deletion by selections didn't update the `times_used` variable (For all grids).
- fixed cel tilemaps having wrong `times_used` when loading project
- Fix Tilemap in place only mode appearing empty when loading project


OverloadedOrama (approved):
Looks good to me. I couldn't find any issues while testing, so it should be good to get merged now. Thank you!

PR #1258, opened 2025-06-01
Added area fill mode where regions from the merging of all layers is filled

Variable-ind:
https://github.com/user-attachments/assets/4d710c03-6602-4ed8-aab6-e8326bce12a4


OverloadedOrama (approved):
Thank you!

PR #1262, opened 2025-06-04
Expose onion ignoring as a layer property

Variable-ind:
Now that we have layer properties, i figured i should move this there as well.

https://github.com/user-attachments/assets/5d52ff72-2461-479c-bd5b-44720629fbb0


OverloadedOrama (approved):
Thank you!

Issue #1250, opened 2025-05-24
Bad performance when drawing inside Group Layer that Blend Mode != Pass Trough


AngeleAkane:
**Pixelorama version:**

v1.1.2-dev

**Where did you download Pixelorama from?**

Nightly

**OS/device including version:**

Windows 10

**Issue description:**

When drawing inside a group layer (either inside the layer, which does not allow anything or in any layer inside) in performance it is considerably reduced even if it is only a single layer.

I do not know if this is normal, I consider that not since Blend Mode does not affect the performance of any other individual layer. Even if they are several layers with several blend modes on top of each other inside or outside Group Layers with Pass Trough.

Note that this only applied when drawing inside or in the group layer directly, and not when drawn outside it. 
In addition, you will always need at least one layer inside.

I discovered it while experimenting with the effects of Layer on FX button. I thought that the bad performance was affected by the effects, even being deactivated. But no, I put the group in "Blend Mode: normal" and that made these slower layers.

Of course, that does not mean that the effects also reduce performance but it seems understandable because they are applied in real time. But I don't think the groups should affect this having in count that normal layers are not affected like this.

**Steps to reproduce:**
1. Create a Group Layer
2. Put AT LEAST ONE Layer inside the Group Layer
3. Change the Group Layer's Blend Mode to ANYTHING BUT Pass Trough (Normal, Darken, Multiply...)
4. Draw, move or do something in any layer INSIDE the Group Layer or try to draw in the Group Layer DIRECTLY.
5. Enjoy lowered FPS.
* This is not applied when drawing outside the group.

Variable-ind:
This is a known issue but unfortunately it can't be fixed 😔.
It occurs for some systems (e.g mine doesn't reduce performance much). The passthrough mode was actually introduced as a workaround for this very purpose.

AngeleAkane:
> This is a known issue but unfortunately it can't be fixed 😔. It occurs for some systems (e.g mine doesn't reduce performance much). The passthrough mode was actually introduced as a workaround for this very purpose.

Damn. So ... do I close the issue or leave it open to remind us that it is an un-fixeable problem?

Variable-ind:
@AngeleAkane I tried to improve performance a little bit, how is it now?
PR: https://github.com/Orama-Interactive/Pixelorama/pull/1269
Download (it'd be fine but remember to make backups just in case 😁): https://nightly.link/Orama-Interactive/Pixelorama/actions/runs/15584027369

[...]

PR #1269, opened 2025-06-11
Boost performance of groups with blend_mode != PassThrough

Variable-ind:
NOTE: This is a performance PR so this would need a little testing, also it should be merged *after* Pixelorama 1.1.2 gets released (just a preference, not a requirement)

Attempt at solving https://github.com/Orama-Interactive/Pixelorama/issues/1250. Boosts the performance of GroupLayers to roughly twice (47% to be precise) it's current state by introducing cache.
- In consecutive generations, when the shader is same, calling `RenderingServer.material_set_shader` again and again (in `generate_image()`) takes up unnecessary computation time (Which makes up most of the 47% time i talked about -- per generation). There is now a way to reuse the material in last generation for the next one.

- when a GroupLayer gets blended, if it has another GroupLayer as child, even if the contents of child GroupLayer didn't change, it ends up being re-blended regardless. To fix this there is now a cache which save time if the contents of a GroupLayer didn't change.

As it uses cache this takes up a little more memory but it's not noticeable.

### DIFF
--- a/src/Classes/Layers/GroupLayer.gd
+++ b/src/Classes/Layers/GroupLayer.gd
@@ -64,14 +64,16 @@ func blend_children(frame: Frame, origin := Vector2i.ZERO, apply_effects := true
 			"origin_x_positive": origin.x > 0,
 			"origin_y_positive": origin.y > 0,
 		}
-		var c_key := [
-			_cache_texture_data, metadata_image.get_data(), origin.x > 0, origin.y > 0
-		]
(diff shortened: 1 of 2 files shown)


OverloadedOrama (approved):
I tested it a bit and it works well. I think it should be fine to get merged for v1.1.2, it's a very welcome performance boost and it doesn't seem to be breaking anything. Thank you very much!

Issue #1282, opened 2025-06-21
Some keyboard shortcuts broken on non-QWERTY layouts


zibetnu:
**Pixelorama version:**
v1.1.1-stable

**Where did you download Pixelorama from?**
Steam

**OS/device including version:**
A desktop with Linux Mint 22.1 Cinnamon

**Issue description:**
The mixed use of physical and Latin keycodes breaks shortcuts in some layouts. On Workman, for example, pressing B will select the text tool instead of the bucket tool.

**Steps to reproduce:**
1. Use Workman layout
2. Press B in an attempt to select the bucket tool
3. See that the text tool was selected instead

PR #1283, opened 2025-06-21
Fix broken tool shortcuts on some keyboard layouts

zibetnu:
The mixed use of physical and Latin-equivalent keycodes for tool shortcuts made some layouts have multiple shortcuts on the same key. Only one of those shortcuts would be used.

Resolves: #1282


OverloadedOrama (approved):
Looks good to me, thank you!

PR #1299, opened 2025-07-06
Rewrite backup system

Variable-ind:
The new system keeps backups of upto 20 last sessions (older session backups get removed)
This is basically the LifeSaver extension, integrated to core.

(some other related fixes i found while integrating):
- fixed `Global.popup_error(tr("File failed to save. Error code %s (%s)") % [err, error_string(err)])` accidentally placed after return instead of before.
- removed duplicate occurrences of `"user://backups"` to a single variable (which should be less prone to bugs)


OverloadedOrama (approved):
Looks good to me, thank you!

PR #1300, opened 2025-07-11
Palette QOL changes

Variable-ind:
I noticed some aspects of the palette panel were in need of improvement so this PR addresses these
- Clicking a swatch with Left/Right color now directly add the color to the desired swatch. This saves the trouble of re-arranging the swatch (if required) to the desired place each time a new color is added
- Previously when deleting a color from palette, you'd have to select color and then press delete button. This is fine for removing one or two colors but when removing a chunk of colors, this task becomes cumbersome. **You can now remove colors by `Ctrl + click` on the desired swatch.**

## Before
https://github.com/user-attachments/assets/f6effdac-e3ea-4797-8666-b5357cd8c554

## After
https://github.com/user-attachments/assets/e8da0fa1-c616-4db2-8afc-f4ae305124c0

- A new `Auto add colors` option has been made available. When enabled, **NEW** colors drawn on canvas will automatically get added to the palette if space is available. (This does not remove colors)

### DIFF
--- a/src/Autoload/Palettes.gd
+++ b/src/Autoload/Palettes.gd
@@ -17,6 +17,7 @@ var palettes_write_path := Global.home_data_directory.path_join("Palettes")
 var palettes: Dictionary[String, Palette] = {}
 ## Currently displayed palette
 var current_palette: Palette = null
+var auto_add_colors := false
 
 # Indexes of colors that are selected in palette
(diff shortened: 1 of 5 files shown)


OverloadedOrama (approved):
Looks good, thank you!

Issue #1297, opened 2025-07-02
Web-page displays as garble in Chinese PC


kinggary92:
**Pixelorama version:**

v1.1.2

**Where did you download Pixelorama from?**

GitHub Release

**OS/device including version:**

windows 10

**Issue description:**

Either official or self-built (godot exports as HTML5) web-page displays as garble. May happen to all computers in Chinese/Japanese/Korean system language. Garble turns to normal English character after I force `locale='en'` in `index.html` as GPT recommended, and then in perference settings it can be seen that only Chinese/J/K options are garbled.

**Steps to reproduce:**
Find a Chinese PC and try office web-page.

OverloadedOrama:
Thanks for reporting, closing as duplicate of #1279

Issue #1279, opened 2025-06-19
The Chinese character became the tofu.


Genouka:
**Pixelorama version:Web Version on https://www.oramainteractive.com/Pixelorama/**

**Where did you download Pixelorama from?**
Web

**OS/device including version:**
Android System WebView 131.0.6778.260

**Issue description:**

The Chinese character became the tofu.

![Image](https://github.com/user-attachments/assets/81b4a5ee-9dc8-4598-b28b-0cb5a10a0e78)

**Steps to reproduce:**
Just open the website.

JarmyGame:
<img width="809" height="539" alt="Image" src="https://github.com/user-attachments/assets/379c7b19-0c52-45b6-8cab-331d0e7021b0" />

Still present, and it affects Japanese and Korean too — not just Chinese.

In Preferences → Language, `[ja_JP]` and `[ko_KR]` render as tofu boxes while all Latin and Cyrillic entries display correctly. Selecting `[ja_JP]` then leaves the UI in English rather than switching, so the missing glyphs may also be preventing the locale from applying.

Worth noting: the translations themselves are complete. `ja_JP.po` is 139,711 bytes — the largest of all 65 files in `Translations/`, where untranslated locales sit at ~102,000. So finished Japanese, Korean and Chinese translations exist and simply never reach web build users.

Environment: Chrome 152.0.7977.77 on Windows 11 (10.0.26200)

PR #1303, opened 2025-07-15
Added fallback fonts for chinese/japanese/korean

LordArugula:
I added NotoSans fonts for Chinese (simplified and traditional), Japanese, and Korean as fallback fonts. The Roboto font doesn't have glyphs for these languages. 

Closes https://github.com/Orama-Interactive/Pixelorama/issues/1297 and https://github.com/Orama-Interactive/Pixelorama/issues/1279


OverloadedOrama:
Thanks for the contribution! To be honest with you, I am not a huge fan of this solution. This adds an extra 26MB for fonts that most people will not sure, and especially on the Web platform which is the one that has the issue, this is going to add a lot of extra loading time. The proper solution would be for fallback fonts to be supported on the Web platform in Godot and the browsers' side, just like they are for the rest of the platforms. https://github.com/godotengine/godot/issues/78921

Until that is supported, a better solution may be to create extensions that add these fonts, or to allow users to load their own fonts to use them for the user interface and the text tool, which is something I'd like to do anyway at some point.

PR #1304, opened 2025-07-15
boost speed of bucket in the pressence of selection by about 87%

Variable-ind:
This introduces two independent optimizations:
## 1.  `lock_selection_rect` optimization:
Stores existing selection rect as cache and forces `get_selection_rect` in `SelectionMap` to use cache instead. use this to save time when the rect isn't meant to change between repeated calls (e.g see it's use in Bucket tool `_flood_fill`). This causes roughly an 87% boost to the area fill algorithm

## 2.  area_fill compatible with selection size:
when segments were created in `_compute_segments_for_image`
```
for j in image.get_height():
	_add_new_segment(j)
```
When we have selection, this ends up adding segments not necessary for calculation. As an optimization, i change it to
```
var y_range = [0, image.get_height()]
	_area_end_idx = project.size.y - 1
	_area_start_idx = 0
	if project.has_selection:
		var selection_rect := project.selection_map.get_selection_rect(project)
		_area_start_idx = selection_rect.position.y
		_area_end_idx = selection_rect.end.y - 1
		y_range = [selection_rect.position.y, selection_rect.end.y]
	for j in range(y_range[0], y_range[1]):
		_add_new_segment(j)
```
This dramatically decreases the fill time even further.

### Benchmark of area_fill on a `30x30` selection in a `646x646` canvas
| Version | Area fill time (msec) |
| --- | --- |
| `master` | 3834 |
| `PR` | 32 |

### Related optimization that tagged along.
- The **lock_selection_rect** optimization is used by `draw_tool` and `draw_fill_gap` as well (this will be noticeable when drawing a line using **Pencil** on a 5000x5000 canvas in presence of selection).

### DIFF
--- a/src/Classes/SelectionMap.gd
+++ b/src/Classes/SelectionMap.gd
@@ -4,6 +4,9 @@ extends Image
 const INVERT_SHADER := preload("res://src/Shaders/Effects/Invert.gdshader")
 const OUTLINE_INLINE_SHADER := preload("res://src/Shaders/Effects/OutlineInline.gdshader")
 
+## An optimization technique
+var _selection_rect_cache := Rect2i()
+
 
 func is_pixel_selected(
 	pixel: Vector2i, calculate_offset := true, project := Global.current_project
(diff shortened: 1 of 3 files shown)


OverloadedOrama (approved):
Looks good to me, thank you!

PR #1306, opened 2025-07-20
further optimize area fill

Variable-ind:
- in `_color_segments()` i used `fill_rect()` instead of filling pixel by pixel (and then use `image.convert_rgb_to_indexed()` at the end). This reduces fill time by about half

- Empty images and selections skip the unnecessary calculations and immediately start filling themselves.

I don't think this breaks something but testing for index mode is required nonetheless :grin:


OverloadedOrama (approved):
Thanks!

PR #1311, opened 2025-07-28
Recalculate start and end points if user deliberately changed the frame during animation.

Variable-ind:
Implements:
> While viewing the animation of some frames under a specific tag (after having checked the animation plays only on frames of the same tag), if you press a frame of another tag you view it for a fraction of a second and then it goes back to showing you the frames of the initial tag. It would be better if it continued playing the animation of the second tag, starting from the frame you just clicked, instead of the first one, since the user clicked the other frame for a reason.

from https://github.com/Orama-Interactive/Pixelorama/discussions/587#discussioncomment-1812691


OverloadedOrama (approved):
Thank you!

PR #1312, opened 2025-07-28
Added ability to easily set tag time

Variable-ind:
You can now adjust the time taken by frames in a tag **without** disturbing their relative duration ratio.

https://github.com/user-attachments/assets/bb59723e-30e9-4011-a2f8-1c6e56a46ec5


OverloadedOrama (approved):
Thanks!

PR #1327, opened 2025-08-18
vertical zoom (via Codex CLI)

ballerburg9005:
On many older systems like CPC or C64, pixels are twice as wide. So artists can't use programs with mismatched pixel scaling.

To address this problem when using Pixelorama, I used GPT-5 to code vertical zoom and briefly tested it.

It seems to be a simple thing that works well.

Please double-check if it makes sense.


ballerburg9005:
vertical2

PR #1328, opened 2025-08-18
trivial bugfix Godot 4.4.1 version update

LazerVV:
complains without


OverloadedOrama:
Thanks for the contribution! Can you explain what the bug is and what Godot is complaining about? I'm also using Godot 4.4.1 and I haven't noticed any bug related to this.

LazerVV:
I am not really sure, ChatGPT guessed first that it was due to Godot 4.4. 

But I think it was actually caused by updating Pixelfed and older files in the config folder.

In any case it makes sense to me to check if the keys are actually present before accessing them.

PR #1340, opened 2025-08-28
Various fixes related to grid snapping and tilemap layers

Variable-ind:
THOROUGH testing is required
- Fixed snapping to grid when pasting
- Clear selection when using move tool and Tool.is_placing_tiles() is `true`
- Fixed tiles not updating when pasting and Tool.is_placing_tiles() is `true`
- Fixed errors popping up (editor only) when bound is zero


OverloadedOrama (approved):
Should be good to go, I didn't encounter any issues while testing. Thanks!

PR #1343, opened 2025-09-02
When removing the palette using undo, it now selects the palette whose name closely matches the removed palette

Variable-ind:


OverloadedOrama (approved):
Thanks!

PR #1372, opened 2025-11-20
Fix id mismatch when extension adds multiple items in the same menu.

Variable-ind:
With the default id = -1


OverloadedOrama (approved):
Thanks!

PR #1383, opened 2025-11-28
QOL improvements to project properties UI

Variable-ind:
- In the `NewTileMapLayerDialog`, use place holder text as tileset name if the field is left empty.
- Added ability to search tilesets in project properties dialog.
- Added ability to rename tilesets in project properties dialog.
- Add first tile as preview if it is found
- Changed labeling to be double lined
- Fixed project name allowed to be a non-valid name through Project properties

https://github.com/user-attachments/assets/38f1f988-0937-4163-8813-66143af603d0


OverloadedOrama (approved):
Thank you!

PR #1384, opened 2025-11-29
Fixed some harmless errors (but still pollute the log file)

Variable-ind:
-  'Attempting to make child window exclusive'
Error appears when importing files through File > open
likely because we do not wait for file dialog to fully close before importing. which causes two exclusive windows open at the same time

- 'Invalid color code: #'
Happens when opening extension explorer. because it tries to convert it to a color first, which is not intended


OverloadedOrama (approved):
Looks good to me, thank you!

PR #1385, opened 2025-11-29
QOL Changes to import options for Spritesheet

Variable-ind:
- Use a ValueSliderV2 instead of spinboxes.
- Added a preset system what auto fills with valid "square" sizes possible for the imported spritesheet.
- Allow excluding/Including empty tiles from spritesheet.
- Revised old "Spritesheet as layer" algorithms. They now use one for loop instead of two :tada:
- Fixed `frame_indices` variable (Inside Spritesheet as layer algorithms) being set the wrong value

Screenshot_2025-11-29_16-57-01


OverloadedOrama (approved):
Thank you!

PR #1413, opened 2025-12-28
Implement undo/redo for layer properties

### LINKED ISSUE #1261: Changes to layer properties can't be undone, aren't treated as changes to the file
**Pixelorama version:** v1.1.1-stable

GitHub Releases

**Issue description:**
Various changes can be made to layers that aren't undoable and don't get counted as changes to the file.
"Various changes" includes changing a layer's name, visibility, opacity, lock, etc. and the values in Layer Properties other than blend mode.

Not being able to undo a changed name or opacity is a nuisance (particularly when it undoes something elsewhere, which can be hard to notice) but I think the more significant result of this is that the file isn't marked as dirty/changed, so the file can be closed without being prompted that there are unsaved changes.

**Steps to reproduce:**
- Open a new file
  - (or any file that has no unsaved changes)
- Rename `Layer 1` to `PLayer won` or something that amuses you
- Note that the `(*)` doesn't get appended to the tab/title bar.
- Note that the file/app can be closed without prompting (and if the file exists, the new layer name is not preserved)

Bartkk0:
Reopening since I accidentally closed #1412

Fixes #1261

### DIFF
--- a/Translations/Translations.pot
+++ b/Translations/Translations.pot
@@ -2911,6 +2911,48 @@ msgstr ""
 msgid "Redo: Change frame duration"
 msgstr ""
 
+msgid "Undo: Rename Layer"
+msgstr ""
+
+msgid "Redo: Rename Layer"
+msgstr ""
+
+msgid "Undo: Change Layer Opacity"
+msgstr ""
+
+msgid "Redo: Change Layer Opacity"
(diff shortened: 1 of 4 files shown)


OverloadedOrama (approved):
Looks good to me, thank you!

PR #1417, opened 2026-01-02
Add support for animated layer effects and a keyframe-based timeline

OverloadedOrama:
Adds a keyframe-based view to the timeline, allowing users to animate layer effects with support for interpolation. In the future this timeline can also be used for other stuff such as animating 3D objects and bones (#1314). Not too sure about the UI, I especially don't really like how you switch between the two modes, but for now it's good enough and we'll make improvements later. This will be merged in a separate branch meant for version 1.2, and once it's good enough, all of the changes from that branch will be merged to master.

https://github.com/user-attachments/assets/834a044d-5b40-426f-be63-14b2c534b50e

The keyframe-based timeline works by seeing all of the layer effects of the currently selected layer, and creating tracks for every property of every effect that can be animated. Each layer effect contains an `animated_params` Dictionary, which is a Dictionary containing another Dictionary that maps the frame indices (int) to another Dictionary of value, trans, ease and a layer-scope unique ID.
Example:
```
{
	"offset":
	{
		0: {id: 0, "value": Vector2(0, 0), "trans": 0, "ease": 2},
		10: {id: 1, "value": Vector2(64, 64), "trans": 1, "ease": 3},
	},
	"wrap_around":
	{
		1: {id: 2, "value": false, "trans": 0, "ease": 0},
		3: {id: 3, "value": true, "trans": 0, "ease": 0},
		10: {id: 4, "value": false, "trans": 0, "ease": 0},
	},
}
```

### DIFF
--- a/src/Classes/LayerEffect.gd
+++ b/src/Classes/LayerEffect.gd
@@ -9,14 +9,22 @@ var layer: BaseLayer
 var category := ""
 var params: Dictionary[String, Variant] = {}
 ## A Dictionary containing another Dictionary that
-## maps the frame indices (int) to another Dictionary of value, trans and ease.
+## maps the frame indices (int) to another Dictionary of value, trans, ease,
+## and a layer-scope unique ID.
 ## Example:
 ## [codeblock]
-##{"offset":
+##{
+##	"offset":
 ##	{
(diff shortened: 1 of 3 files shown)


Zireael07:
Wow, that must've been a lot of work as Godot's built-in AnimationPlayer control isn't exposed

PR #1420, opened 2026-01-04
Keyframe Animation QOL inprovements

Variable-ind:
- Fix keyframe getting created at the wrong frame when the uses right clicks at the same 'x-position' as the frame marking

- Hide extra UI when no FX are present


OverloadedOrama (approved):
Thanks!

PR #1422, opened 2026-01-05
QOL improvements when adding or selecting keyframes

Variable-ind:
- Newly added keyframe gets selected automatically (it won't be selected when user undos and then redos)
- Better box selection. Previously, checked if the top-left corner of keyframe was in the box, and ended up sometimes not selecting the keyframes in cases similar to what is shown below

a
b


Variable-ind:
wrong branch

PR #1423, opened 2026-01-05
 QOL improvements when adding or selecting keyframes

Variable-ind:
- To move single keyframe, you no longer have to select it first, just click and drag (it will be selected when drag is released)

https://github.com/user-attachments/assets/e0ffb654-acf1-48b1-a15c-275e586eec8a

- Newly added keyframe gets selected automatically (though it won't be selected when user undos and then redos)
- Snap the cursor to the frame, as we don't have any useful behavior defined for cursor being in the middle of frames, this makes more sense
- Better box selection. Previously, checked if the top-left corner of keyframe was in the box, and ended up sometimes not selecting the keyframes in cases similar to what is shown below

a
b


OverloadedOrama (approved):
Thanks!

PR #1426, opened 2026-01-07
Fix right tool not selected by stylus when invert button is pressed

Variable-ind:


OverloadedOrama (approved):
Looks good to me, thank you!

PR #1432, opened 2026-01-23
Fixed various visual bugs during undo/redo

Variable-ind:
## Bug 1
- make an outline effect on layer 1
- make an outline effect on layer 2
- select layer 2 and undo once, the outline on layer 2 is removed AND canvas is updated. Do an undo once again (layer 2 still selected) this time effect on layer 1 is indeed shown as removed but canvas is NOT updated.

## Bug 2
- Make 2 layers, both visible and having content
- while having only one layer selected, toggle the global visibility button. the canvas only correctly updates when you switch layers


OverloadedOrama (approved):
Thanks!

PR #1455, opened 2026-03-02
Add new image size presets for various aspect ratios

InsaneAwesomeTony:
added more presets to the new canvas/image options. holy shit git commits and pull requests are difficult.


InsaneAwesomeTony:
omfg i finally did it. it was way harder than it should have been.

PR #1456, opened 2026-03-08
Added ability to split spritesheets per layer

Variable-ind:
Implementation of https://github.com/Orama-Interactive/Pixelorama/discussions/846


OverloadedOrama (approved):
Thank you!

PR #1468, opened 2026-03-23
Allow decimals for reference image values

makinori:
I believe if you scale the project, the scale of reference images can go into a decimal. I somehow caused this value into 13.34 for my project and have already finished hours of drawing.

I adjusted the panel to show decimals and set snapping enabled by default as to keep the original functionality. Holding CTRL whilst dragging will allow for 0.01 precision.

image


makinori:
I should also mention, when importing a new reference image in my project, the scale gets set to 6.25. This PR would allow this to be indicated and adjusted which is helpful.

PR #1472, opened 2026-03-27
Use a centralized crash monitor

Variable-ind:
Pixelorama previously had a separate crash detection systems for extensions and sessions. The crash detection system for extensions was more advanced in the sense that it was designed to have a much smaller probability give false positives in cases where multiple instances were involved.

This PR generalizes/centralizes that crash detector to do both tasks (detecting crashes through sessions, and singling out faulty extensions)

As always testing is appreciated :grin:

### DIFF
--- a/src/Main.gd
+++ b/src/Main.gd
@@ -19,6 +19,7 @@ var splash_dialog: AcceptDialog:
 			splash_dialog = load(SPLASH_DIALOG_SCENE_PATH).instantiate()
 			add_child(splash_dialog)
 		return splash_dialog
+var _last_session_last_project := ""
 
 @onready var top_menu_container := $MenuAndUI/TopMenuContainer as Panel
 @onready var main_ui := $MenuAndUI/UI/DockableContainer as DockableContainer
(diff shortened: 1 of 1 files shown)


Variable-ind:
> when it launches another instance, I receive the message "Pixelorama did not close properly last time. Would you like to restore a previous version". Despite the fact that the first instance did not close at all. Quite annoying too :-)

@DaveInDev could you download this nightly build to see if this bug is fixed (Make sure that `Open last project on startup` is **turned OFF** in `Edit > Prefferences > Startup`):
https://nightly.link/Orama-Interactive/Pixelorama/actions/runs/23671980353

DaveInDev:
hi @Variable-ind , thanks for your help.

on build [23671980353], with "Open last project on startup turned OFF", there is no more error message, but a new Pixelorama instance is launched for any subsequent double-clicked file.

Note that when "Open last project on startup turned ON", there is still no error message, too, but still the double opening of the same file. I'll check the other build [23680356806]

Variable-ind:
> Note that when "Open last project on startup turned ON", there is still no error message, too, but still the double opening of the same file. I'll check the other build [23680356806]

yea, that is fixed by the other build.
these 2 PRs addresses the issues separately.

Variable-ind:
@DaveInDev As the other PR is already merged in master, i just updated this PR here is the new build (this contains solution to both issues):
https://nightly.link/Orama-Interactive/Pixelorama/actions/runs/23692681907

PR #1473, opened 2026-03-27
Fix project opened twice when open last project is enabled

### LINKED ISSUE #1278: Windows "open with..." opens a PXO file twice.
**Pixelorama version:**
1.1.1 stable (but also tried 1.1.1 rc1)

github official

Windows 10 x64 up to date

**Issue description:**
I windows, I linked the opening of a "PXO" file with pixelorama.exe, but then, when I double-click on a PXO file in windows explorer, it opens the selected file twice in Pixelorama.

**Steps to reproduce:**
In windows explorer, right-click on a valid PXO file, use "open with..." and select pixelorama.exe .
-> pixelorama opens with two tabs containing the same PXO file.

It would be nice to open the file only once. It would be also nice that if Pixelorama is already opened, then this instance would be used to open any subsequent files.

Variable-ind:
Fixes #1278

### DIFF
--- a/src/Main.gd
+++ b/src/Main.gd
@@ -544,8 +544,8 @@ func get_last_project_path() -> String:
 
 
 func load_last_project(using_previous_session := false) -> void:
-	# NOTE: When projects are loaded through CLI, the last_project_path gets overriden, and we
-	# have to pass an override path in that senario.
+	# NOTE: When projects are loaded through CLI, the last_project_path gets overridden, and we
+	# have to pass an override path in that scenario.
 	if OS.get_name() == "Web":
 		return
 	# Check if any project was saved or opened last time


Variable-ind:
> Windows "open with..." opens a PXO file twice.

@DaveInDev could you download this nightly build to see if this bug is fixed (Make sure that `Open last project on startup` is **turned ON** in `Edit > Prefferences > Startup`):
https://nightly.link/Orama-Interactive/Pixelorama/actions/runs/23680356806

DaveInDev:
hi @Variable-ind,  thanks for your help.

On this build, with Open last project on startup turned ON or OFF, a double-clicked file is opened only once. Great.
But there is still the error message "Pixelorama did not close properly last time. Would you like to restore a previous version", either with option ON or OFF.

If every double-clicked file was opened in a single instance of the program, could it solve these problems ?

Variable-ind:
> hi @Variable-ind, thanks for your help.
> 
> On this build, with Open last project on startup turned ON or OFF, a double-clicked file is opened only once. Great. But there is still the error message "Pixelorama did not close properly last time. Would you like to restore a previous version", either with option ON or OFF.
> 
> If every double-clicked file was opened in a single instance of the program, could it solve these problems ?

Both problems can be fixed at the same time, the way it works is that i have 2 pull requests:
- one fixes the error issue
- the other fixed the file opening twice issue
Both Pull requests solve each issue separately and are combined after merging to master

This is what the end result should look like: https://nightly.link/Orama-Interactive/Pixelorama/actions/runs/23692681907

PR #1474, opened 2026-03-27
Intersect and Logical AND blend modes

Variable-ind:
- Intersect blend mode crops the result to the common area between blend and the base layers.
- Logical AND blend mode performs Logical AND operation on colors between the blend and the base layers (pixels containing same colors survive and pixels containing different colors are erased).

## Additional Fixes:
- Fix Eraser blend mode being unpredictable at alpha < 1

Update 1: This should be thoroughly tested (I'm not joking).
Update 2: This should be relatively stable now (testing is still appreciated but i'm pretty confident now).


Variable-ind:
A demonstration of power

https://github.com/user-attachments/assets/7d71e1fe-c502-4cf1-bc5b-e8b592a377e7

OverloadedOrama (approved):
Thank you!

PR #1476, opened 2026-03-29
GPU usage improvement

DeriLULZZ2:
i had realized some time ago that pixelorama was consuming an high amount of gpu usage, i thought that it was strange because in the optimization part of the application settings did something that claimed to decrease gpu usage, after sometime i decided to investigate and, aparently pixelorama still redrawed the whole screen when the window is unfocused, i fixed that by just faking one screen and hiding the actual game screen, now it consumes one ABSURDLY lower amount of gpu:
before: Captura de imagem_20260329_165338 now: Captura de imagem_20260329_165443

> (Pixelorama is desguised as godot for godot reasons)
> (Sorry for my bad english)
> (It has 10 commits because i had changed some things that i was'nt suppost to)

### DIFF
--- a/src/UI/Canvas/Selection.gd
+++ b/src/UI/Canvas/Selection.gd
@@ -43,6 +43,7 @@ func _input(event: InputEvent) -> void:
 
 
 func _draw() -> void:
+	marching_ants_outline.visible = Global.current_project.has_selection
 	transformation_handles.queue_redraw()
 
 


OverloadedOrama:
Hello, thank you for your contribution and for bringing attention to the GPU usage issue. I did some testing to find out what is causing the screen to redraw constantly even when Pixelorama is idle, and based from what I see using Godot's Visual Profiler, the CPU & GPU usage remains unchanged when the selection's marching ants outline Sprite2D is invisible. So I pushed a commit that simply hides that Sprite2D when there's no active selection (4a2417ef4e779e5f7199b844ac40f20a6892591d). Can you give it a try to see if the issue still persists? Thanks!

DeriLULZZ2:
ok, apparently that commit fixed the issue

PR #1488, opened 2026-04-20
Add the ability to exclude certain layers properties from undo redo

Variable-ind:
Adds a way to exclude/include certain timeline actions from the undo/redo stack. The prefferences are added in the `Timeline` section. currently they are for changes of:
- blend modes (Change through API)
- layer locking
- layer visibility
- layer opacity (Change through API)
- cel opacity (Change through API)

I am done with this pr for now, further options when required can be added in later PR's.

## Note worthy addition:
- Cel opacity now works with undo/redo.


OverloadedOrama (approved):
Thanks!

PR #1494, opened 2026-04-22
Added RGB color shifting & fixed HSV shifting

Variable-ind:
there were 2 issues:

- Replaced HSV <-> RGB algorithms by ones used by Godot (in it's source code) 

- Use integers in the shader. this makes it consistent between Layer FX and Image effect versions.

- The saturation and value were being incremented very awkwardly (i assume by percentages), so i changed them to how hue was being shifted.

## New Feature:
Added RGB shifting to brightness dialog


OverloadedOrama (approved):
Thanks!

PR #1497, opened 2026-04-29
Fix Extension Exporters not working.

Variable-ind:
Fixes by adding a temporary id as the extension of the exporter added by Extensions.


OverloadedOrama (approved):
Thanks!

PR #1501, opened 2026-05-01
Bone layers keyframe

Variable-ind:
Continuation of #1314

This PR also modifies the keyframe system in order to be compatible with the new layer type


Variable-ind:
Continued in https://github.com/Orama-Interactive/Pixelorama/pull/1571

PR #1507, opened 2026-05-15
Allow setting project size according to clipboard

Variable-ind:
implements https://github.com/Orama-Interactive/Pixelorama/discussions/1505


OverloadedOrama (approved):
Looks good to me, thank you!

PR #1508, opened 2026-05-15
Add an option to repeat animation N times

Variable-ind:
Implements https://github.com/Orama-Interactive/Pixelorama/discussions/1506

Should be good but testing is appreciated


OverloadedOrama (approved):
Thanks!

PR #1516, opened 2026-06-03
Add custom callback functions for opening files with a custom extension

AlRado:
Add custom callback functions for opening files with a custom extension.
This makes them very easy to use in extensions:
```
func _enter_tree() -> void:
extension_api = get_node_or_null("/root/ExtensionsApi")
extension_api.import.add_open_callback("scr", Callable(self, "open_scr_file"))

func _exit_tree() -> void:
extension_api.import.remove_open_callback("scr")

func open_scr_file(path: String) -> void:
# Here's the implementation of the image loading logic in a custom format
```


OverloadedOrama (approved):
Looks good to me, thank you!

PR #1523, opened 2026-06-12
Fixed darkening of edges in the gaussian blur effect

Variable-ind:
Suggested by a user on discord (SOS1), the credit should go to them.

I have went through the code, my analysis is:
- The fix is basically the usage of `unshaded` render mode and then pre-multiplying alpha (e.g `blur_color_sum += sample_col.rgb * sample_col.a`) and then taking alpha-weighted average of the rgb color at the end.
- The weights (e.g `float w = exp(-i * i * 0.5);`) are just added for smoothing and don't contribute to the fix.


OverloadedOrama (approved):
Thank you!

PR #1532, opened 2026-06-20
Bugfix: Split layers breaks selected/visible layers

SjamesE:
As shown in this screenshot, the "selected layers" from the Layers dropdown is not working when split layers is checked:
Screenshot 2026-06-20 122611

Fixed:
image


OverloadedOrama (approved):
Looks good to me, thank you very much!

PR #1535, opened 2026-06-22
Add ability to connect to project/cel updaters through API

Variable-ind:
NOTE: The API level has been updated to 9

This also makes sure that the updater signals always come from the "current" project (i-e the logic to smartly re-connect the signals whenever the current Cel or Project is changed is done through the API).

This is the evolved and more generalized form of the `signal_current_cel_texture_changed()` method in API. For compatibility reasons, i have let that method remain and added a deprecation warning to it. 

# Demo Extension
```
extends Node

@onready var extension_api: Node  ## A variable for easy reference to the Api

## This script acts as a setup for the extension
func _enter_tree() -> void:
	# NOTE: Use get_node_or_null("/root/ExtensionsApi") to access api.
	# NOTE: See https://www.oramainteractive.com/Pixelorama-Docs/extension_system/extension_api for
	# detailed documentation.
	extension_api = get_node_or_null("/root/ExtensionsApi")

	# |==== Your code goes here ====|
	extension_api.signals.current_project_signal(updater_layers, "layers_updated")
	extension_api.signals.current_cel_signal(updater_texture, "texture_changed")
	# |=============================|


func updater_layers():
	print("Layers of %s are changed" % extension_api.project.current_project.name)


func updater_texture():
	var project = extension_api.project.current_project
	print(
		"Texture changed of Project (%s) Layer: %s, Frame %s."
		% [project.name, str(project.current_layer + 1), str(project.current_frame + 1)]
	)


## Gets called when the extension is being disabled or uninstalled (while enabled).
func _exit_tree() -> void:
	# Remember to remove things that you added using this extension
	# Disconnect any signals and queue_free() any nodes that got added.
	extension_api.signals.current_project_signal(updater_layers, "layers_updated", true)
	extension_api.signals.current_cel_signal(updater_layers, "layers_updated", true)
```


OverloadedOrama (approved):
Thanks!

PR #1539, opened 2026-06-29
Cleaned the AnimatableObject class, removed shadow variables in BaseLayer

Variable-ind:
Follow up of 7b0d8b8b127c6dc386a84574958f2e279610ae65

- Added documentation comments to `params`, `get_params()` and `get_animated_property()`
- Added an `unset_keyframe()` method to `AnimatableObject`. This cleans up the `add_effect_keyframe()` method in `KeyframeTimeline.gd` and give more flexibility in case we want to use it for other types (e.g adding a bone keyframe etc.)
- Renamed `params` in BaseLayer's `display_effects()` as they were being marked as shadow variables.
- Removed `static` from `get_animated_property()`. I don't think it's used anywhere (as `AnimatableObject.get_animated_property()`) and i doubt we will ever use it like that, as this class is meant to be inherited by other classes so they can already use `get_animated_property()` without having the need for it to be static (and needing unnecessary parameters)
- In `get_animated_property`, changed:
```
if not is_interpolatable_type(min_value):
			to_return[param] = max_value
			return to_return
```
- To:
```
if not is_interpolatable_type(min_value):
			return min_value
```
- I thought returning `max_value` was a typo as in this situation we should have treated it like the `TRANS_CONSTANT` case

### DIFF
--- a/src/Classes/AnimatableObject.gd
+++ b/src/Classes/AnimatableObject.gd
@@ -5,7 +5,17 @@ signal keyframe_set
 
 const TRANS_CONSTANT := -1
 
+## These are the default values for the animation calculator to fall back on when a keyframe is
+## not found.
+## Example:
+## [codeblock]
+##{
+##	"offset": Vector2(0, 0),
+##	"wrap_around": false,
+##}
+## [/codeblock]
 var params: Dictionary[String, Variant] = {}
+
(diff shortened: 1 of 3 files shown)


OverloadedOrama (approved):
Looks good to me, thank you!

PR #1541, opened 2026-06-30
Added option to use Project name and Project path as defaults for export

Variable-ind:
Also moved FFMPEG to the Export section in preferences


Variable-ind:
The Pr is outdated now

PR #1544, opened 2026-06-30
Use the Command key for timeline selection on macOS

### LINKED ISSUE #1313: Can not selecting multiple cels on macOS
**Pixelorama version:**
v1.1.2

itch.io

macOS 15.5 M1

**Issue description:**
Can not select multiple individual cels/frames/layers, holding the `Control` key and left-clicking equals right-click on macOS.

For macOS, we might want to follow Apple's GUI guidelines and use the `Command` key with left-clicking for individual select/deselect.

Arthurmtro:
Closes #1313

## Issue
On macOS it is impossible to multi-select cels, frames or layers in the timeline. Holding **Control** and left-clicking is interpreted by macOS as a right-click, so instead of adding the item to the selection, the context menu opens. The same applies to **Control + drag** for swapping cels/frames/layers.

## Cause
All of these interactions polled the `ctrl` input action:
```gdscript
elif Input.is_action_pressed("ctrl"):
```
Because the OS converts Control + left-click into a right-click, this branch is never reached on macOS.

## Fix
Following Apple's interface guidelines, macOS should use the **Command** key for individual selection. Command + click stays a normal left-click, so it reaches the selection logic correctly.

Added a small helper to `Global`:
```gdscript
func is_ctrl_or_cmd_pressed() -> bool:
	if OS.get_name() == "macOS":
		return Input.is_key_pressed(KEY_META)
	return Input.is_action_pressed(&"ctrl")
```
and routed the timeline buttons (`CelButton`, `FrameButton`, `LayerButton`, `LayerMainButton`) through it, for both multi-selection and swap-on-drag. Behaviour on Windows and Linux is unchanged.

## Testing
- `gdformat --check` and `gdlint` pass on the changed files.
- The project compiles with no new errors introduced by these changes.
- Manual runtime testing on macOS is in progress; I will confirm here once Command + click multi-selection and Command + drag swapping are validated in a running build. Feedback from other macOS users is welcome.

If maintainers prefer using `is_command_or_control_pressed()` via `gui_input` instead of the `Global` helper, I am happy to adjust.

### DIFF
--- a/src/Autoload/Global.gd
+++ b/src/Autoload/Global.gd
@@ -872,6 +872,15 @@ func is_linux_or_bsd() -> bool:
 	return OS.get_name() in ["Linux", "FreeBSD", "NetBSD", "OpenBSD", "BSD"]
 
 
+## Returns [code]true[/code] if the platform's selection modifier is held down.
+## On macOS this is the Command key, because Control+left-click is interpreted by the
+## operating system as a right-click; on every other platform it is the Control key.
+func is_ctrl_or_cmd_pressed() -> bool:
+	if OS.get_name() == "macOS":
+		return Input.is_key_pressed(KEY_META)
+	return Input.is_action_pressed(&"ctrl")
+
+
 ## Print this when Pixelorama launches so it can be stored in the log files.
 ## This info may help us better debug issues certain users may have.
 func get_system_info() -> String:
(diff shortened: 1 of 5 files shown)


OverloadedOrama (approved):
Looks good to me. Thank you very much!

PR #1545, opened 2026-07-02
When pasting with grid center snapping enabled, make the center of the selection fall below the center of the grid

Variable-ind:


OverloadedOrama (approved):
Thank you!

PR #1562, opened 2026-07-29
Persist export dialog settings per-project

### LINKED ISSUE #1503: Spritesheet export settings are not saved
**Pixelorama version:**

1.1.10

Flatpak

Ubuntu 24.04

**Issue description:**

When exporting an image as spritesheet, it is possible to select different export options e.g. number of lines. But these settings are not saved in .pxo. Therefor i always have to edit them again (after closing and opening Pixelorama).
These settings should be saved, so export is consistent for all sessions and can be reproduced when sharing .pxo

Qmaker-programmer:
Export dialog settings (spritesheet orientation, lines count, frame range, resize, crop mode, split layers, etc.) used to live only in the Export autoload's global state, so they were shared across all open projects and lost whenever Pixelorama was closed and reopened.

Move these settings onto Project itself, serialize/deserialize them with the rest of the project data (like export_directory_path and file_format already were), and fully sync the export dialog's UI from the current project's settings when it is opened.

Fixes #1503

Assisted by Claude Sonnet 5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,7 @@ Built using Godot 4.6.3
 - Fixed hardware related issue where layer opacity & blend mode did not update in certain mobile GPUs. [#1546](https://github.com/Orama-Interactive/Pixelorama/issues/1546)
 - Make UI elements in the audio layer properties stack properly.
 - Fixed extension exporters not working. [#1497](https://github.com/Orama-Interactive/Pixelorama/pull/1497)
+- Export dialog settings (spritesheet layout, frame range, resize, crop mode, etc.) are now saved per-project and restored when the project is reopened. [#1503](https://github.com/Orama-Interactive/Pixelorama/issues/1503)
 
 ### Removed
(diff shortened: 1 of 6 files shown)


Qmaker-programmer:
**Formatting fix**

The `static-checks` job was failing because `gdformat --diff .` flagged 1 file needing reformatting: [`[src/Autoload/Export.gd](https://claude.ai/epitaxy/src/Autoload/Export.gd)`](src/Autoload/Export.gd).

The new `export_save_quality`, `export_number_of_digits`, and `export_interpolation` fields (added in this PR to persist export settings) pushed three calls (`save_jpg`, `pad_zeros`, `resize`) past `gdformat`'s line-length limit.

Fixed by running `gdformat` on the file, which split those calls across multiple lines. Verified locally:
- `gdformat --diff .` → 247/247 files unchanged
- `gdlint src/Autoload/Export.gd` → no problems found

Commit: b4cb36ad

OverloadedOrama:
Thanks for the contribution, superseded by #1574

PR #1569, opened 2026-08-01
Fix quick tool buttons triggering when a multi state tool is in use

Variable-ind:
Fixes the bug for:
- Curve tool
- Polygon select tool
- Isometric box tool


OverloadedOrama (approved):
Thank you!

PR #1574, opened 2026-08-04
Make each project have their own export profile

### LINKED ISSUE #1503: Spritesheet export settings are not saved

Variable-ind:
Added an ExportProfile subclass in `Export` Autoload, each project has it's own copy of it and gets saved/loaded with the project. All export related variables have been moved to this sub-class

Testing is required (mostly just verification that i didn't miss something)

Supersedes PR: https://github.com/Orama-Interactive/Pixelorama/pull/1562
Fixes: https://github.com/Orama-Interactive/Pixelorama/issues/1503

### DIFF
--- a/src/Autoload/Export.gd
+++ b/src/Autoload/Export.gd
@@ -121,7 +121,6 @@ class ExportProfile:
 			"new_dir_for_each_frame_tag": var_to_str(new_dir_for_each_frame_tag),
 			"number_of_digits": var_to_str(number_of_digits),
 			"separator_character": var_to_str(separator_character)
-
 		}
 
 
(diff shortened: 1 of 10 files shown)


OverloadedOrama (approved):
Looks good to me, thank you!

PR #1576, opened 2026-08-06
Turn off redundant  flags for windows with  flag enabled

Variable-ind:
Needed because of https://github.com/godotengine/godot/issues/117698.


OverloadedOrama (approved):
Thanks!

PR #1577, opened 2026-08-07
Fix text applied by text tool not having correct transparent colors

### LINKED ISSUE #1556: drop shadow bleeding otherer colors when using text tool
**Pixelorama version:**
1.1.10-stable linux

steam

Linux Mint, Geforce 4090

**Issue description:**
The drop shadow on a layer bleeds other colors when using the text tool.

In the added screenshots you can see two rows of text. The one on top has been added using an older version of Pixelorama. The one below has been added with the new version of Pixelorama. The shadow is bleeding other colors into shadow. I think it has more to do with the text insert function.

Image

Image
Image

**Steps to reproduce:**
Open up Pixelorama and add a text to a layer. Add Drop Shadow.

Variable-ind:
Fixes: https://github.com/Orama-Interactive/Pixelorama/issues/1556

## Additional information
Instead of rendering text directly on cel image, this PR does it on a separate image and then blends it with the cel's image. This makes godot auto resolve the Colors whose alpha was zero but had non-zero RGB components.

This Also ends up fixing the following issues as well:

- The original issue of colors bleeding in drop shadow

- Colors getting distorted when new text was written such that it overlapped with some previous text

- Boxes appeared in background of text when it was applied (similar to the issue on previous line)

- Some transparent areas were not filled when bucket tool was applied after writing text on a new layer

No

### DIFF
--- a/src/Tools/UtilityTools/Text.gd
+++ b/src/Tools/UtilityTools/Text.gd
@@ -119,7 +119,9 @@ func text_to_pixels() -> void:
 
 	var undo_data := _get_undo_data()
 	var project := Global.current_project
-	var image := project.frames[project.current_frame].cels[project.current_layer].get_image()
+	var cel_image := project.frames[project.current_frame].cels[project.current_layer].get_image()
+	if cel_image == null:
+		return
 
 	var vp := RenderingServer.viewport_create()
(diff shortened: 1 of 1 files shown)


OverloadedOrama (approved):
Looks good to me, thank you!

PR #1579, opened 2026-08-09
Fix LCD antialiasing for text tool and make text crop to selection bounds

Variable-ind:

- The text now respects the selection area.
- Did a small regression from https://github.com/Orama-Interactive/Pixelorama/pull/1577. It turns out that for `FONT_ANTIALIASING_LCD`, we actually need background information for it to work, so we add it back only in case for this mode. Luckily this shouldn't cause transparency issues as i observed `FONT_ANTIALIASING_LCD` mode does not have transparent pixels anyway.
    - I Should note that when FONT_ANTIALIASING_LCD is used over transparent area, the text will be placed inside a black box, this isn't a bug as the antialiasing is working independent of alpha it sees `(0, 0, 0, 0) -> Transparent` as `(0, 0, 0) -> Black`, The reason it shows without black box in preview is because it's seeing the checker as a color (gray/white), instead of as transparent (See video for illustration).

https://github.com/user-attachments/assets/10768fb7-6ba0-4a94-9adf-7850662c6a28

No


OverloadedOrama (approved):
Thanks!

PR #1580, opened 2026-08-12
A more powerful Index Map

Variable-ind:
This PR adds more features to the Index map effect that are now possible due to the key frame timeline.
- Added a `Use pixel position for transparent colors`. When enabled, Index map effect uses the "Canvas position" for look up in "Map Texture" instead of the R,G values of the color.
- You can now wrap the coordinates if offset causes them out of Map texture's area.
- You can now also set some offset to the look up coordinates i-e (X, Y) + Offset.
- Scaling can also be gone using the index map.
- Added a checkbox that stops the process at the intermediate stage and returns the image with R and G components equal to the X and Y coordinates of the color to be used for lookup (After applying all the offset/wrap/zoom calculations). This is meant to be used in cases where we want to use the index map and not want to draw indices by hand.

## Video
https://github.com/user-attachments/assets/5373bfb1-5ec6-40b8-9d54-59a36073be9e

## Demo Project
[Demo.zip](https://github.com/user-attachments/files/31037235/Demo.zip)

No

## Status: Ready for review

### DIFF
--- a/src/Classes/AnimatableObject.gd
+++ b/src/Classes/AnimatableObject.gd
@@ -206,6 +206,11 @@ func deserialize(dict: Dictionary) -> void:
 		animated_params = str_to_var(dict["animated_params"])
 
 
+## preserves compatibility for name changes across different version.
+func is_param_valid(_param_name: StringName) -> bool:
+	return true
+
+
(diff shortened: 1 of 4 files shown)


OverloadedOrama (approved):
Looks great, thank you!

PR #1587, opened 2026-08-17
Allow smoothly moving values when sliding even when the bounds have reached.

Variable-ind:
One of the pain points for me regarding value sliders is that they are very limiting after the bounds are reached. YES i can manually type in greater or smaller numbers but it's very destructive and slow to workflow (Especially when setting negative values for vector2 and vector3 sliders in shader fx).

The solution i have come up with is to allow sliders to behave like how it is handles in blender. The slider is now still responsive to dragging even when the bar has reached bounds. For cases when users want to immediately move to a smaller value they would have to type the value in the slider or drag it back, but the benefits in return is well worth it.

No


OverloadedOrama (approved):
Thanks!

PR #1588, opened 2026-08-17
Use _unhandled_input for tool switching

### LINKED ISSUE #1581: New Quick Tools Feature problem
**Pixelorama version:** 1.2 and 1.2.1

**Where did you download Pixelorama from?** Steam

**OS/device including version:** Win 11 Pro latest

**Issue description:** A quick tool shortcut activate temporarily a toll if no other tool is active.
Release works if it is activated.

**Steps to reproduce:**
Press the eraser shortcut (i set it to H for the test) than the eraser is active as long as i hold H. After release H it swap back to the Pencil i.e..
Now the problem: If i wanna eraser something while i paint with the pencil next i press H and it happens nothing.
So the Quick shortcut seem not to work while working active with another tool. To activate the quick tool you have to release the left mousebotton or hold the pen away from tablet.
In short: just try to use it. ;)
As a tip, the "swap tool" for the right and left mousebutton works like a charm.
This behaviour is needed for the new quick tools feature.

Thanks alot

Variable-ind:

Fixes https://github.com/Orama-Interactive/Pixelorama/issues/1581, instead of using `Input.is_action_pressed(&"activate_left_tool")` and `Input.is_action_pressed(&"activate_right_tool")` to see if tool is being used it now only gets called when the shortcut for the quick tool does not conflict with any `shape_perfect, shape_center, shape_displace` shortcuts.

The original method was mainly to prevent color picker getting selected when we were pressing Alt, this new method does the same thing in essence

No

### DIFF
--- a/src/Autoload/Tools.gd
+++ b/src/Autoload/Tools.gd
@@ -685,7 +685,7 @@ func is_placing_tiles() -> bool:
 	return Global.current_project.get_current_cel() is CelTileMap and TileSetPanel.placing_tiles
 
 
-func has_selection_tool() -> bool:
+func selection_tool_has_selection() -> bool:
 	if not Global.current_project.has_selection:
 		return false
 	for mouse_button in _slots:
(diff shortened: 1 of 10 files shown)


OverloadedOrama (approved):
Thank you!

PR #1594, opened 2026-08-26
integer zoom now zooms exponentially like the normal zoom

Variable-ind:
Implements #1592. In current implementation the integer zoom mode appeared to be slower the more you zoomed in, this fixes that issue and the zoom feels more natural.

### Before

https://github.com/user-attachments/assets/ade88c2f-606b-4fe9-8a66-7d1a9f3becf1

### After

https://github.com/user-attachments/assets/e748c2c0-27c9-4f96-b58e-04327c05ca50

### Won't this skip over many zoom values that user might need?
Short answer is no. It does skip values but they are mostly those that users will likely never need. for example user definitely needs zooms like 100% 200% --- 1000%. What they likely never need is zoom like `12400%`, `12500%` `12600%` `12700%` (They all look similar and can be skipped and just show one of them, then move on). Users can still type zoom levels if they want extra precision

No


OverloadedOrama (approved):
Thank you!

PR #1603, opened 2026-08-29
Implementing exporting projects as Aseprite files

Variable-ind:
Adds export option for Aseprite (tested on 1.3.18.3-dev, commit [e537f0b043b3246ffa0f784d46494322f5b63a9b](https://github.com/aseprite/aseprite/commit/e537f0b043b3246ffa0f784d46494322f5b63a9b)). This is basically a reverse process of the AsepriteParser class

## Additional information
This only does a best effort conversion and doesn't include features used by Pixelorama but not in Aseprite
- [x] Project info (FPS etc...)
- [x] Cel linking
- [x] Project palettes
- [x] Cel, Layer, Tag Userdata.
- [x] Blend modes
- [x] Fix Indexed mode quirks
- [x] Tilemaps

No


OverloadedOrama (approved):
Looks good to me. I didn't do extensive testing, but it seems to be working well, and most of the feedback will be gathered from people that will use it, so I think it should be good to be merged now and we can fix any issues later. Thank you!

